PR #3, opened 2025-11-21
fix: carplay template button update race condition

g4rb4g3:
This PR fixes the race condition when starting the app and bringing up the trip selector immediately on CarPlay.
Most often the trip selector back button was missing and other buttons that should not have been shown did show up...
This PR makes sure template updates happen on the main thread as the @MainActor annotation on CPTemplate's implies.

### DIFF
--- a/packages/react-native-autoplay/ios/templates/MapTemplate.swift
+++ b/packages/react-native-autoplay/ios/templates/MapTemplate.swift
@@ -182,9 +182,7 @@ class MapTemplate: NSObject, AutoPlayTemplate, AutoPlayHeaderProviding,
 
         template.tripEstimateStyle = isDark ? .dark : .light
 
-        DispatchQueue.main.async {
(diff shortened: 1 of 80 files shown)


packages/react-native-autoplay/ios/templates/MapTemplate.swift
```diff
@@ -186,9 +180,11 @@ class MapTemplate: NSObject, AutoPlayTemplate, AutoPlayHeaderProviding,
             isDark ? .dark : .light
         )
 
-        invalidate()
-
         template.tripEstimateStyle = isDark ? .dark : .light
+
+        DispatchQueue.main.async {
+            self.invalidate()
```
SamuelBrucksch: Invalidate is annotated with main actor and we call it without the dispatch in other places. Is this required?

SamuelBrucksch (approved):

PR #15, opened 2026-01-08
Refactor header actions handling in MapTemplate

LukasB-DEV:
# Fix: Ensure ActionStrip is always set for NavigationTemplate

## Problem
Android Auto's `NavigationTemplate` requires an `ActionStrip` to be set, otherwise it throws an `IllegalStateException` during build:

```
java.lang.IllegalStateException: Action strip for this template must be set
at androidx.car.app.navigation.model.NavigationTemplate$Builder.build
```

This crash occurred when `config.headerActions` was `null` or not provided, as the previous implementation would skip calling `setActionStrip()` entirely.

As ActionStrip needs at least one Action as a fallback the app icon is rendered

```
java.lang.IllegalStateException: Action strip must contain at least one action
```

## Solution
- Always call `setActionStrip()`, even when no header actions are configured
- Provide an empty `ActionStrip` as a fallback when `config.headerActions` is null
- Add logging to help debug cases where the fallback is used

## Changes
**Before:**
```kotlin
config.headerActions?.let { headerActions ->
    setActionStrip(Parser.parseMapHeaderActions(context, headerActions))
}
```

**After:**
```kotlin
val headerActionStrip = config.headerActions?.let { headerActions ->
    Parser.parseMapHeaderActions(context, headerActions)
} ?: run {
    Log.w("MapTemplate", "No header actions provided, using empty ActionStrip")
    ActionStrip.Builder().build()
}
setActionStrip(headerActionStrip)
```

## Testing
- Verified that templates without header actions no longer crash
- Confirmed that templates with header actions continue to work as expected

### DIFF
--- a/packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/MapTemplate.kt
+++ b/packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/MapTemplate.kt
@@ -74,23 +74,15 @@ class MapTemplate(
             config.mapButtons?.let { buttons ->
                 setMapActionStrip(Parser.parseMapActions(context, buttons))
             }
-            val headerActionStrip = config.headerActions?.let { headerActions ->
-                Parser.parseMapHeaderActions(context, headerActions)
+                        config.headerActions?.let { headerActions ->
+                setActionStrip(Parser.parseMapHeaderActions(context, headerActions))
             } ?: run {
-                ActionStrip.Builder()
-                    .addAction(
-                        Action.Builder()
-                            .setIcon(
-                                CarIcon.Builder(IconCompat.createWithResource(context, R.drawable.ic_launcher))
-                                    .build()
-                            )
-                            .setOnClickListener { /* No-op, just satisfies requirement */ }
-                            .build()
-                    )
-                    .build()
+                setActionStrip(
+                    ActionStrip.Builder()
+                        .addAction(Action.APP_ICON)
+                        .build()
+                )
             }
-            setActionStrip(headerActionStrip)
-            
             val travelEstimates =
                 if (config.visibleTravelEstimate == VisibleTravelEstimate.FIRST) destinationTravelEstimates.firstOrNull() else destinationTravelEstimates.lastOrNull()
             travelEstimates?.let {
--- a/packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/MapTemplate.kt
+++ b/packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/MapTemplate.kt
@@ -77,8 +77,18 @@ class MapTemplate(
             val headerActionStrip = config.headerActions?.let { headerActions ->
                 Parser.parseMapHeaderActions(context, headerActions)
             } ?: run {
-                Log.w("MapTemplate", "No header actions provided, using empty ActionStrip")
-                ActionStrip.Builder().build()
+                Log.w("MapTemplate", "No header actions provided, using app icon as fallback")
+                ActionStrip.Builder()
+                    .addAction(
+                        Action.Builder()
+                            .setIcon(
+                                CarIcon.Builder(IconCompat.createWithResource(context, R.drawable.ic_launcher))
+                                    .build()
+                            )
+                            .setOnClickListener { /* No-op, just satisfies requirement */ }
+                            .build()
+                    )
+                    .build()
             }
             setActionStrip(headerActionStrip)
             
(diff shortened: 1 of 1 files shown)


packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/MapTemplate.kt
```diff
@@ -73,9 +74,24 @@ class MapTemplate(
             config.mapButtons?.let { buttons ->
                 setMapActionStrip(Parser.parseMapActions(context, buttons))
             }
-            config.headerActions?.let { headerActions ->
-                setActionStrip(Parser.parseMapHeaderActions(context, headerActions))
+            val headerActionStrip = config.headerActions?.let { headerActions ->
+                Parser.parseMapHeaderActions(context, headerActions)
+            } ?: run {
+                Log.w("MapTemplate", "No header actions provided, using app icon as fallback")
```
g4rb4g3:
```suggestion
```
We don't want to log these :)

g4rb4g3:
Congrats on the first "external" PR on this repo, so far only Iternio related people did them. 🥳 
Great catch, so far I wasn't considering this as we have at least a single button all the time on our app.
Thank you very much for this, I just have a minor change proposal on removing the logging.

g4rb4g3 (changes requested):

packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/MapTemplate.kt
LukasB-DEV: Logging is removed :)

packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/MapTemplate.kt
```diff
@@ -73,9 +74,23 @@ class MapTemplate(
...
+                                CarIcon.Builder(IconCompat.createWithResource(context, R.drawable.ic_launcher))
+                                    .build()
+                            )
+                            .setOnClickListener { /* No-op, just satisfies requirement */ }
+                            .build()
+                    )
+                    .build()
             }
+            setActionStrip(headerActionStrip)
+
```
g4rb4g3:
```suggestion
                        config.headerActions?.let { headerActions ->
                setActionStrip(Parser.parseMapHeaderActions(context, headerActions))
            } ?: run {
                setActionStrip(
                    ActionStrip.Builder()
                        .addAction(Action.APP_ICON)
                        .build()
                )
            }
```

g4rb4g3: I just checked out your branch and realized that this is not working, the imports are missing and R can't be imported since R.drawable.ic_launcher belongs to the app and can't be accessed on the lib.
There is actually a default Android Auto action for handling this which seems to work out fine. Keep in mind that the icon is converted to black and white by AA so this might look terrible in the end...
image

g4rb4g3 (changes requested):

packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/MapTemplate.kt
LukasB-DEV: do you have a other idea what to display?

LukasB-DEV:
@g4rb4g3 do you have another idea what to display beside the app icon?

g4rb4g3:
No I guess this is fine as fallback, in the end developers should specify a proper icon imho. Since material icons are bundles there are quite a few options available anyways.

packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/MapTemplate.kt
```diff
@@ -73,8 +74,14 @@ class MapTemplate(
             config.mapButtons?.let { buttons ->
                 setMapActionStrip(Parser.parseMapActions(context, buttons))
             }
-            config.headerActions?.let { headerActions ->
+                        config.headerActions?.let { headerActions ->
```
g4rb4g3:
```suggestion
            config.headerActions?.let { headerActions ->
```

g4rb4g3 (changes requested):

g4rb4g3 (approved):

SamuelBrucksch (approved):

SamuelBrucksch:
Thank you.

SamuelBrucksch:
I just released version 0.1.13 that includes your changes:
https://www.npmjs.com/package/@iternio/react-native-auto-play/v/0.1.13

PR #20, opened 2026-01-09
Support both LEGACY and DEFAULT for UIManagerType

SamuelBrucksch:
Ensure backwards compatibility with old React Native versions.


g4rb4g3 (approved):
Thank God we'll move to new arch soon and can remove this then. 😆

PR #22, opened 2026-01-11
fix: Fix nil crash in SceneStore.getRootTraitCollection when root scene isn’t ready

LukasB-DEV:
Bug: SceneStore.getRootTraitCollection() force‑unwraps a nil scene.
Symptom: App crashes on CarPlay connect with Fatal error: Unexpectedly found nil while unwrapping an Optional value.
("ReactNativeAutoPlay/SceneStore.swift:88: Fatal error: Unexpectedly found nil while unwrapping an Optional value")
Fix: Return default UITraitCollection() if root scene is missing.

### DIFF
--- a/packages/react-native-autoplay/ios/scenes/SceneStore.swift
+++ b/packages/react-native-autoplay/ios/scenes/SceneStore.swift
@@ -85,6 +85,9 @@ class SceneStore {
     }
 
     static func getRootTraitCollection() -> UITraitCollection {
-        return store[SceneStore.rootModuleName]!.traitCollection
+        guard let scene = store[SceneStore.rootModuleName] else {
+            return UITraitCollection()
+        }
+        return scene.traitCollection
     }
 }


g4rb4g3:
Hi @LukasB-DEV, thanks for opening up this one.
Is there any way you can provide an example on how to reproduce this issue?
So far I did not spot it and I wonder how the main scene could not be ready but templates being pushed?
Are you trying to push templates before the connect callback was invoked?

LukasB-DEV:
> Hi @LukasB-DEV, thanks for opening up this one. Is there any way you can provide an example on how to reproduce this issue? So far I did not spot it and I wonder how the main scene could not be ready but templates being pushed? Are you trying to push templates before the connect callback was invoked?

Hi, this happened to me when I quickly connected, disconnected, and reconnected CarPlay. So definitely a edge case, but definitely something that can happen to users ( Layer 8 issue ;) )

g4rb4g3:
Thing is that with a fresh `UITraitCollection` you'll have a 50% chance of not having the proper userInterfaceStyle leading to some wrong colors in the end.
Following the default flow of waiting for the didConnect callback being invoked and pushing templates then should not cause this. It is always CarPlay first initializing the `HeadUnitSceneDelegate` which puts it in the scene store. Only then the didConnect callback is invoked.
I prefer to find the root cause, so in case you have a working reproduction I can check it sometime later.

LukasB-DEV:
@g4rb4g3 Thanks for the reply. In my app I only call setRootTemplate inside didConnect (and guard with isConnected). The crash still happens when users quickly disconnect/reconnect CarPlay, so didConnect fires, setRootTemplate runs, but before template.invalidate executes, the root scene has already been removed. That’s why SceneStore.getRootTraitCollection unwraps nil. Sentry shows this during HybridAutoPlay.setRootTemplate and setTemplateHeaderActions.

Repro we’ve seen:
1) Connect CarPlay and start navigation.
2) Quickly unplug and replug within ~1–2s.
3) App crashes with SceneStore.swift:88.

Proposed fix:
Return traitCollection from root scene if available, otherwise fall back to the window scene (SceneStore.windowSceneModuleName) or any existing scene; only if none exist return UITraitCollection(). This avoids the crash while still preserving correct UI style in most cases.

g4rb4g3:
@LukasB-DEV What is your setup for testing this? Do you see the issue on the example app too? If not could you try to modify it so it resembles your flow and brings up the same error on your side?
Right now I can't reproduce the issue, tested it with the example app on an iPhone 11 (18.7.2) and an iPhone 15 Pro (26.2) and the CarPlay Simulator.
Checking our Sentry does not show errors like yours either.

LukasB-DEV:
@g4rb4g3 In my app I call setRootTemplate + setHeaderActions right on didConnect and then continuously update maneuvers/estimates while navigation is active. If CarPlay disconnects between setRootTemplate and the async invalidate(), the root scene is already removed and getRootTraitCollection unwraps nil.

To reproduce in the example app, add a small delayed update after didConnect, e.g. setTimeout(() => mapTemplate.setHeaderActions(...), 500), then quickly unplug/replug. The timeout fires after disconnect and triggers the crash.

g4rb4g3:
@LukasB-DEV Interesting, so this issue happens only while disconnecting, right? Connecting again has the proper flow again and the scene in place... In theory your approach would not cause any visible issues then.

Still I wonder why you call `setHeaderActions` just after calling `setRootTemplate`. You can specify the `headerrActions` on your `MapTemplateConfig` without calling `setHeaderActions`. What is your reason for calling `setHeaderActions`?

LukasB-DEV:
@g4rb4g3 Yes, the crash only happens during disconnect (race between disconnect and a pending template update). Reconnect afterwards is fine.

I call `setHeaderActions` right after `setRootTemplate` because my header actions are dynamic. The `MapTemplateConfig` headerActions are static at init, but I need to toggle actions based on current navigation state (e.g. show “Stop navigation” only while navigating) and current entitlement (search button only for premium users). Also, if navigation is already active when CarPlay connects, I want the header to reflect that immediately. That’s why I call `setHeaderActions` after setting the root template and whenever nav state changes.

g4rb4g3:
@LukasB-DEV 
Ok, sounds like you could determine that header state before creating the template?
I still looked at this and spent some time on doing a cleanup, maybe you can give it a try using this https://github.com/Iternio-Planning-AB/react-native-auto-play/pull/28?

You should get the proper error now when calling a template function when the scene is gone. So you have to handle the rejected promise then.
> Error: sceneNotFound("operation failed, AutoPlayRoot scene not found")

LukasB-DEV:
@g4rb4g3 thanks for your work. Will try it out tomorrow.

LukasB-DEV:
@g4rb4g3 tested it and it seems to work pretty well.

g4rb4g3:
Great news 😀 , thanks for testing. I'll wait for @SamuelBrucksch on the review. Once it's done we'll do a new release.

PR #24, opened 2026-01-13
Catch error "No templates were available to be popped"

SamuelBrucksch:


packages/react-native-autoplay/ios/scenes/AutoPlayInterfaceController.swift
```diff
@@ -96,10 +96,18 @@ class AutoPlayInterfaceController: NSObject, CPInterfaceControllerDelegate {
...
+                animated: animated
+            )
+        } catch let error as NSError
+            where error.domain == "CarPlayErrorDomain" && error.code == -1
+        {
+            // Silently ignore "No templates were available to be popped" error
+        } catch {
+            // Rethrow any other unexpected errors
+            throw error
+        }
```
g4rb4g3:
```suggestion
        if templateIds.count == 0 {
            return templateIds
        }
```

g4rb4g3 (commented):
I'd prefer to not catch exceptions but do a check on the template count that would be popped. If we don't have any we can early return, similar to what we do on the other pop functions.
That should make these exceptions gone and is a cleaner approach imho. 🙂 
What do you think?

g4rb4g3:
btw, the issue can be reproduced by calling popToRoot on the root template.
image
checking the count fixes it 🎉

packages/react-native-autoplay/ios/scenes/AutoPlayInterfaceController.swift
```diff
@@ -96,10 +96,9 @@ class AutoPlayInterfaceController: NSObject, CPInterfaceControllerDelegate {
             templateIds.append(templateId)
         }
 
-        try await interfaceController.popToRootTemplate(
-            animated: animated
-        )
-
+        if templateIds.count == 0 {
+            return templateIds
+        }
```
SamuelBrucksch, line 101:
```suggestion
        }
        try await interfaceController.popToRootTemplate(
            animated: animated
        )
```

packages/react-native-autoplay/ios/scenes/AutoPlayInterfaceController.swift
```diff
@@ -96,10 +96,12 @@ class AutoPlayInterfaceController: NSObject, CPInterfaceControllerDelegate {
             templateIds.append(templateId)
         }
 
+        if templateIds.count == 0 {
+            return templateIds
+        }
         try await interfaceController.popToRootTemplate(
             animated: animated
         )
```
SamuelBrucksch, line 104:
```suggestion
        )

```

SamuelBrucksch:
Sounds like a better approach...

g4rb4g3 (approved):

PR #26, opened 2026-01-16
Feature/automotive telemetry

g4rb4g3:
This PR adds Android Automotive telemetry and permission request support.


apps/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -7,7 +7,6 @@
       android:label="@string/app_name"
       android:icon="@mipmap/ic_launcher"
       android:roundIcon="@mipmap/ic_launcher_round"
-      android:allowBackup="false"
       android:theme="@style/AppTheme"
```
g4rb4g3, line 10: Since Android 12 other approaches are recommended then setting this flag, also we don't care about it on the example app.

packages/react-native-autoplay/android/gradle.properties
```diff
@@ -1,5 +1,5 @@
 ReactNativeAutoPlay_kotlinVersion=2.1.20
-ReactNativeAutoPlay_minSdkVersion=23
+ReactNativeAutoPlay_minSdkVersion=24
```
g4rb4g3, line 2: It was using 24 already, only this file did not get updated.

apps/example/src/App.tsx
```diff
@@ -13,7 +13,7 @@ import {
   actionStopNavigation,
 } from './state/navigationSlice';
 import { useAppDispatch, useAppSelector } from './state/store';
-import { TelemetryView } from './TelemetryView';
+import TelemetryView from './TelemetryView';
```
g4rb4g3, line 16: why not check it yourself 😝

apps/example/src/AutoPlay.tsx
```diff
@@ -21,7 +21,7 @@ import {
   setSelectedTrip,
 } from './state/navigationSlice';
 import { startAppListening } from './state/store';
-import { TelemetryView } from './TelemetryView';
+import TelemetryView from './TelemetryView';
```
g4rb4g3, line 24: why not check it yourself 😝

packages/react-native-autoplay/android/build.gradle
```diff
@@ -124,16 +124,27 @@ android {
           "${project.buildDir}/generated/source/codegen/java"
         ]
       }
+
+      if (getExtOrDefault("isAutomotiveApp") == "true") {
```
g4rb4g3, line 128: tested and works fine

packages/react-native-autoplay/android/build.gradle
```diff
@@ -124,16 +124,27 @@ android {
...
+        java.srcDirs += 'src/automotive/java'
+        manifest.srcFile 'src/automotive/AndroidManifest.xml'
+      } else {
+        java.srcDirs += 'src/auto/java'
```
g4rb4g3, line 132: this is a bs comment.... why would I add the default main directory again?

packages/react-native-autoplay/android/src/main/AndroidManifest.xml
```diff
@@ -13,16 +13,6 @@
...
-    <uses-permission android:name="android.car.permission.CAR_ENERGY" />
-    <uses-permission android:name="android.car.permission.CAR_ENERGY_PORTS" />
-    <uses-permission android:name="android.car.permission.CAR_EXTERIOR_ENVIRONMENT" />
-    <uses-permission android:name="android.car.permission.CAR_IDENTIFICATION" />
-    <uses-permission android:name="android.car.permission.CAR_INFO" />
-    <uses-permission android:name="android.car.permission.CAR_POWERTRAIN" />
-    <uses-permission android:name="android.car.permission.CAR_SPEED" />
-
-
     <!-- Android Auto specific permissions -->
```
g4rb4g3, line 16: removing automotive permissions on regular builds does impact exactly 0 devices.

packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/HybridAndroidAutoTelemetry.kt
```diff
@@ -1,7 +1,45 @@
...
+import com.margelo.nitro.swe.iternio.reactnativeautoplay.template.AutomotivePermissionRequestTemplate
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlin.coroutines.resume
+
 class HybridAndroidAutoTelemetry : HybridAndroidAutoTelemetrySpec() {
-    override fun registerTelemetryListener(callback: (Telemetry?, String?) -> Unit): () -> Unit {
-        return AndroidAutoTelemetryObserver.addListener(callback)
+    override fun registerTelemetryListener(callback: (tlm: Telemetry?) -> Unit): Promise<() -> Unit> {
+        return Promise.async {
+            AndroidTelemetryObserver.addListener(callback)
```
g4rb4g3: good catch, this was not working at all.... 😬

packages/react-native-autoplay/src/types/Telemetry.ts
```diff
@@ -68,4 +83,18 @@ export type Telemetry = {
    * Vehicle information
    */
   vehicle?: VehicleTelemetryItem;
+  /**
+   * one of VehicleGear enum
+   */
+  selectedGear?: NumericTelemetryItem;
```
g4rb4g3, line 89: yeah but we must type it as number here

packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/TelemetryObserver.kt
```diff
@@ -0,0 +1,54 @@
+package com.margelo.nitro.swe.iternio.reactnativeautoplay
+
+import android.os.Handler
+import android.os.HandlerThread
+
+abstract class TelemetryObserver {
+    abstract fun startTelemetryObserver(): Boolean
+    abstract fun stopTelemetryObserver()
+
+    var telemetryCallbacks: MutableList<(Telemetry?) -> Unit> = ArrayList()
```
g4rb4g3: should be accessed by the js thread only but still a good idea

yarn.lock
```diff
@@ -4408,6 +4408,11 @@ react-is@^19.1.0:
   resolved "https://registry.npmjs.org/react-is/-/react-is-19.1.1.tgz"
   integrity sha512-tr41fA15Vn8p4X9ntI+yCyeGSf1TlYaY5vlTZfQmeLBrFo3psOPX6HhTDnFNL9uj3EhP0KAQ80cugCl4b4BERA==
 
+react-native-config@^1.6.1:
+  version "1.6.1"
+  resolved "https://registry.yarnpkg.com/react-native-config/-/react-native-config-1.6.1.tgz#f8428829fda74384484d773fe36820159c559159"
+  integrity sha512-HvKtxr6/Tq3iMdFx5REYZsjCtPi0RxQOMCs15+DqrUPTNFtWHuEuh+zw7fJp+dmuO79YMfdtlsPWIGTHtaXwjg==
+
```
g4rb4g3, line 4415: this is a auto generated file

packages/react-native-autoplay/src/hooks/useAndroidAutoTelemetry.ts
```diff
@@ -58,46 +77,83 @@ export const useAndroidAutoTelemetry = ({
...
-          setPermissionsGranted(true);
-        })
-        .catch((e) => console.error('*** Android Auto telemetry permissions error', e));
+    if (!requestTelemetryPermissions || requiredPermissions.length === 0) {
+      return;
     }
-    return;
-  }, [requestTelemetryPermissions, requiredPermissions]);
+
+    if (permissionsGranted !== false) {
```
g4rb4g3, line 106: that is actually exactly what it should do, we want to request permissions only when they are not granted, so null and true have to be early return here.

packages/react-native-autoplay/android/src/automotive/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/AndroidTelemetryObserver.kt
```diff
@@ -0,0 +1,170 @@
...
+                            val currentGear = getValue<Int>(p0)
+                            val selectedGear =
+                                if (currentGear == VehicleGear.GEAR_PARK || currentGear == VehicleGear.GEAR_NEUTRAL || currentGear == VehicleGear.GEAR_REVERSE) {
+                                    currentGear
+                                } else {
+                                    // we are not interested in the actual gear, like first, second...
+                                    VehicleGear.GEAR_DRIVE
+                                }
+                            telemetryHolder.updateSelectedGear(selectedGear)
+                            // TODO: sendEvent("automotive_gear_changed", getIsDriving(newSelectedGear))
```
g4rb4g3: ⚠️

packages/react-native-autoplay/android/src/automotive/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/AndroidTelemetryObserver.kt
```diff
@@ -0,0 +1,170 @@
...
+    val vehiclePropertyReceiver = object : CarPropertyManager.CarPropertyEventCallback {
+        override fun onChangeEvent(p0: CarPropertyValue<*>?) {
+            try {
+                if (p0?.status == CarPropertyValue.STATUS_AVAILABLE) {
+                    when (p0.propertyId) {
+                        VehiclePropertyIds.GEAR_SELECTION -> {
+                            // this one reports only the basic gears like drive, park....
+                            val selectedGear = getValue<Int>(p0)
+                            telemetryHolder.updateSelectedGear(selectedGear)
+                            // TODO: sendEvent("automotive_gear_changed", getIsDriving(newSelectedGear))
```
g4rb4g3: ⚠️

packages/react-native-autoplay/android/src/automotive/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/AndroidTelemetryObserver.kt
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
+
+        override fun onErrorEvent(p0: Int, p1: Int) {
+            Log.e(TAG, "error receiving vehicle property id $p0 zone: $p1")
+        }
+    }
+
+    override fun startTelemetryObserver(): Boolean {
+        if (isObserverRunning) {
```
SamuelBrucksch, line 138: In Auto Telemetry we check
```
        val carContext =
            AndroidAutoSession.getCarContext(AndroidAutoSession.Companion.ROOT_SESSION)
                ?: throw IllegalArgumentException(
                    "Car context not available, failed to start telemetry"
                )
```
Before we start the listener. Is this not needed on Automotive or was it overly cautious on Auto?

SamuelBrucksch: Here we instead rather assign the car context immediately, when we create the AndroidTelemetryObserrver instance.

SamuelBrucksch (commented):
Looks good so far, but I have one concern about the CarContext when the TelemetryObserver is created. You assume it is always there, but in Auto implementation, we first check and set it before we actually start the listener, which is called from react native side. So that seems to be more safe. But if the CarContext is already always available when the Observer is initialized, we could as well adjust the Auto impl to be less cautious.

packages/react-native-autoplay/android/src/automotive/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/AndroidTelemetryObserver.kt
g4rb4g3: Good point, even though the car context should be available at that point it doesn't hurt to check it properly.

SamuelBrucksch (approved):

PR #28, opened 2026-01-17
Chore/carplay refactor

g4rb4g3:
This PR brings some CarPlay ❤️ :
* validate scene before calling invalidate on templates
* remove static TemplateStore
* clear template store on disconnect
* reduce force unwraps
* adds a "swift:format" command


packages/react-native-autoplay/ios/scenes/HeadUnitSceneDelegate.swift
```diff
@@ -60,10 +60,11 @@ class HeadUnitSceneDelegate: AutoPlayScene, CPTemplateApplicationSceneDelegate {
         HybridAutoPlay.emit(event: .diddisconnect)
         disconnect()
```
SamuelBrucksch: This calls templateStore.disconnect internally, which clears the template store:
```
    func disconnect() {
        NitroSurface.stop(self.window?.rootViewController?.view)
        self.window = nil
        isConnected = false

        templateStore.disconnect()
        SceneStore.removeScene(moduleName: moduleName)
    }
```
So the template below would always be null

SamuelBrucksch (changes requested):
Nice, please check the comment with the disconnect function.

packages/react-native-autoplay/ios/scenes/HeadUnitSceneDelegate.swift
g4rb4g3: Awesome catch, thank you very much! 😃

SamuelBrucksch (approved):

PR #34, opened 2026-01-27
SoE for Android Automotive

g4rb4g3:
This PR brings SoE for Android Automotive.
image

### DIFF
--- a/packages/react-native-autoplay/src/types/Telemetry.ts
+++ b/packages/react-native-autoplay/src/types/Telemetry.ts
@@ -97,4 +97,8 @@ export type Telemetry = {
    */
   evBatteryInstantaneousChargeRate?: NumericTelemetryItem;
   parkingBrakeOn?: BooleanTelemetryItem;
+  /**
+   * Returns the current battery level in Wh
+   */
+  soe?: NumericTelemetryItem;
 };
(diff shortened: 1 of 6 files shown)


packages/react-native-autoplay/src/types/Telemetry.ts
```diff
@@ -97,4 +97,8 @@ export type Telemetry = {
    */
   evBatteryInstantaneousChargeRate?: NumericTelemetryItem;
   parkingBrakeOn?: BooleanTelemetryItem;
+  /**
+   * Returns the current battery level in Wh
+   */
+  soe?: NumericTelemetryItem;
```
SamuelBrucksch, line 103: Looking at the other properties shouldn't this be rather `evRemainingEnergy` or `evBatteryLevel`or so? soe is very ABRP specific I think...

g4rb4g3: SoE is a common term in EV speech, like SoC or SoH so I guess it is fine.

SamuelBrucksch (approved):

packages/react-native-autoplay/src/types/Telemetry.ts
SamuelBrucksch: maybe we should rename batteryLevel to soc and evBatteryInstantaneousChargeRate to power as well then?

g4rb4g3: yeah, maybe someday, when we run out of more important topics 😛

PR #45, opened 2026-02-06
Chore/bump biome 2.3.14

g4rb4g3:
This PR bumps biome and fixes some issues found while doing that.


SamuelBrucksch (approved):

SamuelBrucksch:
We probably should add the biome check to the pipeline...

PR #47, opened 2026-02-06
possible fix for a crash

SamuelBrucksch:
We got this crash on Sentry:
```
0   ABRP                            0x200a10878         closure in MapTemplate.parseMapButtons
1   ABRP                            0x200a0fe80         MapTemplate.parseMapButtons
2   ABRP                            0x200a1115c         MapTemplate._invalidate
3   ABRP                            0x200a3b920         closure in RootModule.withAutoPlayTemplate<T>
4   ABRP                            0x200a03ad4         HybridMapTemplate
```

According to an AI analysis this would be the right fix. But I'm not too familiar with native iOS code, so @g4rb4g3 could you please have a look?

### DIFF
--- a/packages/react-native-autoplay/ios/templates/MapTemplate.swift
+++ b/packages/react-native-autoplay/ios/templates/MapTemplate.swift
@@ -81,7 +81,11 @@ class MapTemplate: AutoPlayHeaderProviding,
     }
 
     func parseMapButtons(mapButtons: [NitroMapButton]) -> [CPMapButton] {
-        return mapButtons.map { button in
+        return mapButtons.map { [weak self] button in
+            // Capture onPress callback directly to avoid issues with button lifecycle
+            let onPress = button.onPress
+            let buttonType = button.type
+            
             if let glyphImage = button.image.glyphImage,
                 let icon = SymbolFont.imageFromNitroImage(
                     image: glyphImage,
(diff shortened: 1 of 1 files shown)


SamuelBrucksch:
AI Response:

The fix addresses two issues:

1. **Strong reference cycle / dangling reference in closures** - The `CPMapButton` closures were capturing `self` and `button` strongly. When `hideTripSelector()` is called, it triggers `invalidate()` which calls `parseMapButtons()`. If the button callbacks (`button.onPress`) reference objects that have been deallocated or are in the process of being invalidated, calling them would crash.

   **Fix**: Changed to `[weak self]` captures and extracted the `onPress` callback and `buttonType` into local variables at closure creation time, avoiding access to the potentially deallocated `button` object later.

2. **Variable shadowing** - The local variable `mapButtons` was shadowing the instance property `self.mapButtons`, which while not directly causing the crash, is confusing and could lead to unexpected behavior.

   **Fix**: Renamed the local variable to `filteredMapButtons` and explicitly used `self.mapButtons` to access the instance property.

The `EXC_BREAKPOINT` crash was likely caused by Swift's runtime detecting an invalid memory access when the closure tried to access `button.onPress?()` after the `NitroMapButton` object had been deallocated. This commonly happens with Nitro hybrid objects that bridge between JavaScript and native code.

g4rb4g3:
Most likely this is related to the root trait collection crash we see. I don't think it is the onPress callback being triggered at the exact same time invalidate is executing causing this crash then. More likely is the trait collection disappearing while disconnecting and invalidate is ongoing because of some action like hideTripSelector().

https://github.com/Iternio-Planning-AB/react-native-auto-play/pull/49

SamuelBrucksch:
Properly fixed with https://github.com/Iternio-Planning-AB/react-native-auto-play/pull/49

PR #51, opened 2026-02-15
Chore/bump react native 0.83

g4rb4g3:
This PR bums react-native to 0.83 and removes legacy architecture support.


packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/VirtualRenderer.kt
```diff
@@ -216,7 +193,7 @@ class VirtualRenderer(
                     val additionalMarginRight =
                         if (stableArea.right == visibleArea.right && visibleArea.right != width) 0 else defaultMargin
                     val additionalMarginTop =
-                        if (visibleArea.top != stableArea.top || (visibleArea.top > 0 && stableArea.top > 0 && visibleArea.right < width)) 0 else defaultMargin
+                        if (visibleArea.top != stableArea.top || (visibleArea.top > 0 && visibleArea.right < width)) 0 else defaultMargin
```
g4rb4g3, line 196: since visibleArea.top equals stableArea.top the second greater 0 check is not required

packages/react-native-autoplay/ios/ReactHelpers/NitroSurface.m
```diff
@@ -19,30 +19,16 @@ + (void)stop:(nullable UIView *)view {
         return;
     }
 
-    if ([view isKindOfClass:[RCTSurfaceHostingProxyRootView class]]) {
+    if ([view isKindOfClass:[RCTSurfaceHostingView class]]) {
```
SamuelBrucksch, line 22: @g4rb4g3 could this cause any issues here? It changed after all...

SamuelBrucksch (dismissed):
Looks good but please check the AI review comment.

packages/react-native-autoplay/ios/ReactHelpers/NitroSurface.m
g4rb4g3: Ah good catch, it doesn't cause any issues right now as `RCTSurfaceHostingProxyRootView` extends `RCTSurfaceHostingView`. But this is definitely something I'll have to fix before we merge this.

PR #52, opened 2026-02-18
prevent crash on disconnect, when scene is removed and code on js side still executed

SamuelBrucksch:
When CP was disconnected, the react code still executed and called native functions, but the scene was already destroyed, so it crashed with error `operation failed, AutoPlayRootScene scene not found`


g4rb4g3:
Closing in favor of https://github.com/Iternio-Planning-AB/react-native-auto-play/pull/54

PR #53, opened 2026-02-23
Use a ref for useMapTemplate

SamuelBrucksch:
**This is a breaking change.** We probably should make a version 0.4 if we want to do it like this.

The issue is, that on disconnect when we have a logic like
```
const mapTemplate = useMapTemplate()

useEffect(() => {
  mapTemplate.showTripSelector()
  return () => {
    mapTemplate.hideTripSelector()
  }
}, 
[mapTemplate, ...otherProps])
```
where we actually want to hide the trip selector based on other props, the mapTemplate would actually not be valid anymore in the cleanup function on unmount. However it is still defined, so calling that action we will run the native code, although it is already in the process of tearing down everything. Right now this would cause an exception, that the scene is not valid anymore.

When using a ref like in this PR we would be able, to immediately reset the state of the mapTemplate, even if the UI wouldn't render anymore. So the useEffect would look like this:
```
const mapTemplateRef = useMapTemplate()

useEffect(() => {
  mapTemplateRef.current?.showTripSelector()
  return () => {
    mapTemplate.current?.hideTripSelector()
  }
}, 
[mapTemplateRef.current, ...otherProps])
```
And the cleanup function wouldn't do anything, as the ref is already set to null when getting there.

This could potentially replace https://github.com/Iternio-Planning-AB/react-native-auto-play/pull/52 as we wouldn't have to check for the connection state on native side. We could however still keep the logic about not throwing when the scene is gone. Not sure though if we actually need it...

### DIFF
--- a/apps/example/src/AutoPlay.tsx
+++ b/apps/example/src/AutoPlay.tsx
@@ -56,7 +56,7 @@ const AutoPlayRoot = (props: RootComponentInitialProps) => {
       return;
     }
 
-    mapTemplate?.showAlert({
+    mapTemplate.current?.showAlert({
       durationMs: 10 * 1000,
       id: 1,
       primaryAction: {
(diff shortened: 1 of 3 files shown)


g4rb4g3:
Closing in favor of https://github.com/Iternio-Planning-AB/react-native-auto-play/pull/54

PR #59, opened 2026-03-17
update tlm value timestamp, even if value stays the same

SamuelBrucksch:
**Previous behaviour**
If the incoming value is the same as the cached value, we ignore it. 

In case of battery level for example it could mean, that the timestamp does not update for quite a while and the value looks old, although it is actually updated (means the car reports the same value over and over again).

**New behaviour**
We now always update the timestamp, even if the value stays the same, so we know that the data is not stale.


g4rb4g3:
We should do the same on automotive too probably?

SamuelBrucksch:
> We should do the same on automotive too probably?

The Holder is shared between both, or not?

g4rb4g3:
Right, my bad. Next time I'll check the PR on my computer and not on my phone. 🙂

SamuelBrucksch:
Tested on AA and Automotive. Both update the timestamp when the same value is sent again.

g4rb4g3 (approved):

PR #61, opened 2026-03-23
fix: get screen manager from first non-cluster session

laptiieff:
This is a fix for the error which I encountered when set `GridTemplate` as a root template on Android Auto
Here is a repository with minimal app which reproduces the error: [link](https://github.com/laptiieff/android-auto-error-app). Steps to reproduce and fix the error (via patch) are described in readme, as well as short screen recordings of before and after fix behaviors


g4rb4g3 (approved):
As we highly focus on navigation usage we have missed that. Thank you very much, highly appreciated! :)

g4rb4g3:
https://github.com/Iternio-Planning-AB/react-native-auto-play/releases/tag/v0.3.6 out now with you fix in place.

PR #62, opened 2026-03-30
fix: defer makeKeyAndVisible to suppress UIKit rendering warning

sbs44:
`WindowApplicationSceneDelegate` transplants the root view controller from the AppDelegate's non-scene window into a new scene-based `UIWindow` and immediately calls `makeKeyAndVisible()`. This triggers a repeated UIKit console warning:

```
[UIKitCore] Rendering a view (0x…, UIWindow) that has not been rendered
at least once requires afterScreenUpdates:YES.
```

The fix defers `makeKeyAndVisible()` to the next run loop iteration via `DispatchQueue.main.async`, giving UIKit time to lay out the transplanted view hierarchy before the first render pass.

Tested with Expo SDK 55 / React Native 0.83 on the Xcode Simulator CarPlay window — warning is eliminated with no change in behavior.


g4rb4g3:
I do not see this issue with the example app, can you please provide a reproducible and more details so I can check it? Thank you!

PR #67, opened 2026-04-08
feat: support remote URL images in asset type

Lucaaa03:
## Summary

Adds support for loading images from remote HTTP/HTTPS URLs in CarPlay and Android Auto templates. Currently, the `asset` image type only supports bundled images via `require()`. This change enables dynamic images (e.g. operator logos, location photos) from remote URLs.

### Usage

```typescript
// Remote URL (new)
image: { image: { uri: 'https://example.com/logo.png' }, type: 'asset' }

// Bundled asset (unchanged)
image: { image: require('./logo.png'), type: 'asset' }
```

## Changes

| File | Platform | Change |
|------|----------|--------|
| `NitroImage.ts` | JS | Detect `http://`/`https://` URIs and bypass `Image.resolveAssetSource()` |
| `Parser.kt` | Android | Parse remote URIs directly; enable Fresco disk+memory caching for remote images |
| `Parser.swift` | iOS | Add `URLSession`-based synchronous loader with `NSCache` in-memory caching |

## Design decisions

- **Synchronous loading** on both platforms to match the existing template rendering flow. Android already uses `DataSources.waitForFinalResult()`; iOS uses `Data(contentsOf:)`.
- **Caching**: Android leverages Fresco's built-in disk+memory cache for remote images. iOS uses an `NSCache` in-memory cache. Local assets retain their existing behavior (no caching).
- **Backward compatible**: No changes to the `AutoImage` type — remote URLs work through the existing `ImageSourcePropType` which already accepts `{ uri: string }`.

## Motivation

Many apps need to display dynamic images in car display templates — for example logos or photos fetched from an API. Currently this requires bundling every possible image as a local asset, which isn't feasible for dynamic content. This change allows passing any HTTPS URL as an image source.

## Testing

Tested on:
- iOS CarPlay Simulator (Xcode)
- Android Auto DHU

Remote HTTPS images load and display correctly in templates, including color tinting for dark/light mode.


g4rb4g3 (changes requested):
Hi and thank you for your PR.

I was considering this when writing the lib but decided against it — relying on remote images without knowing about network availability and speed felt wrong, which is why I went with bundled assets only. That said, I can see the appeal as a feature.

While reviewing I found a few issues I'd like to address before merging:

- **New image type**: I'd prefer a dedicated type for this rather than reusing `asset`. It's neither a glyph nor a bundled asset, and having its own type would also eliminate the `startsWith("https://")` string comparisons scattered across TS, iOS, and Android.
- **HTTP support**: `http://` images don't work on CarPlay even with `NSAllowsArbitraryLoads` set. I'd suggest dropping HTTP support entirely — HTTPS is the standard nowadays and this avoids the ATS complexity.
- **Error handling**: On iOS, a failing URL silently returns `nil`. On Android, Fresco throws, causing the entire template to not render. Having either on a root template could cause a crash loop. We need consistent, graceful fallback behavior on both platforms.
- **No fetch timeout**: Image fetching blocks indefinitely on both platforms. On a slow or unavailable network this will make the app unresponsive.
- **Caching on Android**: There's already a `BitmapCache` (LruCache, 8 MB) that handles caching for all images. Both Fresco's memory and disk cache should be kept disabled for remote images — same as for local assets — and let `BitmapCache` handle it uniformly.
- **No cache size limit on iOS**: The `NSCache` has no `countLimit` or `totalCostLimit` set, unlike the 8 MB cap on Android.

Lucaaa03:
Hi! Thanks for the review. I will address your findings and commit an enhanced version 😄

packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/Parser.kt
```diff
@@ -518,31 +518,69 @@ object Parser {
...
+                    }
+                } finally {
+                    img?.close()
+                    res?.close()
+                    ds.close()
+                }
+            }
+            thread.start()
+            thread.join(10_000L)
+            if (thread.isAlive) {
```
Lucaaa03: The synchronous approach is intentional — `parseAssetImage` is called from the Car App Library's template rendering pipeline which expects synchronous image resolution. The 10s join timeout is a safety net; if it expires, we return `null` (fallback to no image) rather than relying on `interrupt()` to stop the I/O. The thread will complete on its own, the result is simply discarded.

packages/react-native-autoplay/ios/templates/Parser.swift
```diff
@@ -763,27 +763,71 @@ class Parser {
...
+        let config = URLSessionConfiguration.default
+        config.timeoutIntervalForRequest = 10
+        let session = URLSession(configuration: config)
+        var resultData: Data?
+        let semaphore = DispatchSemaphore(value: 0)
+        session.dataTask(with: url) { data, _, _ in
+            resultData = data
+            semaphore.signal()
+        }.resume()
+        _ = semaphore.wait(timeout: .now() + 10)
```
Lucaaa03: `parseAssetImage` is always called on a background thread by the Car App Library's rendering pipeline, never on the main thread — so the semaphore won't cause UI freezes. The synchronous pattern matches Android's Fresco `waitForFinalResult` approach and is required because the rendering pipeline expects images to be resolved synchronously.

Lucaaa03:
I've addressed all 6 points in a separate commit. I'm looking forward to another review 😄

packages/react-native-autoplay/src/types/Image.ts
```diff
@@ -29,4 +29,13 @@ export type AutoImage =
        */
       color?: ThemedColor | string;
       type: 'asset';
+    }
+  | {
```
g4rb4g3, line 33: Good start, please add a matching type on NitroImage.ts so we can get rid of the "https://" check on native side too.
No need to specify placeholder values on NitroImage.convert too then. :)

packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/Parser.kt
```diff
@@ -518,31 +518,69 @@ object Parser {
...
+                        fetchedBitmap = drawable?.toBitmap(width = img.width, height = img.height, Bitmap.Config.ARGB_8888)
+                    }
+                } finally {
+                    img?.close()
+                    res?.close()
+                    ds.close()
+                }
+            }
+            thread.start()
+            thread.join(10_000L)
```
g4rb4g3: 10 seconds seems quite long since this is per request, right?
Maybe add an optional timeoutMs property to the remote image type and use a fallback off 500ms in case it is not specified? Having it as prop allows overriding it in case someone prefers to adjust it.

packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/Parser.kt
```diff
@@ -518,31 +518,69 @@ object Parser {
...
+            var fetchedBitmap: Bitmap? = null
+            val thread = Thread {
+                val ds = try {
+                    Fresco.getImagePipeline().fetchDecodedImage(imageRequest, context)
+                } catch (_: Exception) { return@Thread }
+                val res = try {
+                    DataSources.waitForFinalResult(ds)
+                } catch (_: Exception) { ds.close(); return@Thread }
+                val img = res?.get()
+                try {
```
g4rb4g3: this seems to be almost identical to the bundled image handling, lets unify this in some function.

g4rb4g3 (changes requested):
Nice progress, left a few inline comments.

Lucaaa03:
Thanks for the quick follow-up review! I pushed three commits addressing the three points, rebased onto current `master`:

**1. Matching type on `NitroImage.ts`**
Added a dedicated `RemoteImage` interface and extended the `NitroImage` union to `GlyphImage | AssetImage | RemoteImage`. Regenerated via `yarn specs`, which renames the variant to `Variant_GlyphImage_AssetImage_RemoteImage` across Kotlin/Swift/C++ and adds a `RemoteImage` struct. `NitroImageUtil.convert()` now returns a flat `{ uri, color, timeoutMs }` for the remote case — no more placeholder `height`/`scale`/`packager_asset` fields. The `"https://"` prefix check is gone from both native parsers. The large diff in commit c0811fc9 is `yarn specs` output only, reproducible locally.

**2. Configurable timeout, 500ms default**
Added optional `timeoutMs?: number` to the remote image type. Fallback is `500L` ms on Android (`Thread.join`) and `0.5s` on iOS (`URLSessionConfiguration.timeoutIntervalForRequest` + semaphore wait).

**3. Unified bundled + remote handling**
- Android `Parser.kt`: extracted `buildImageRequest()`, `fetchBitmap(timeoutMs)`, `fetchSync()`, and `applyTintAndCache()`. `parseAssetImage` and `parseRemoteImage` now share the Fresco pipeline and tinting logic — they differ only in URI source and whether a timeout is applied.
- iOS `Parser.swift`: extracted `applyTint()`; both paths share tinting and trait-collection handling.

Split the work so the handwritten changes (ae8cd6c9, 7 files) are reviewable separately from the nitrogen output (c0811fc9). A follow-up commit (16503d06) restores `thread.interrupt()` after the join timeout, which I accidentally dropped during the refactor. Looking forward to your feedback!

g4rb4g3 (approved):
well done, thank you very much!

PR #72, opened 2026-04-17
Feature/voice recording

g4rb4g3:
This PR adds voice recording capabilities. Recordings are passed as raw PCM stream.
It can be used either with Android Auto and CarPlay showing a indication or be used on the mobile device directly.

image
image


packages/react-native-autoplay/ios/utils/VoiceInputManager.swift
```diff
@@ -0,0 +1,224 @@
...
+                self.continuation = nil
+            }
+        }
+
+        return samplesAsData(samples)
+    }
+
+    func stop(interfaceController: AutoPlayInterfaceController? = nil) {
+        guard !isStopping else { return }
+        isStopping = true
```
SamuelBrucksch, line 68: From Cursor:
> The stop() method uses a plain Bool (isStopping) as a reentrancy guard, but it can be called concurrently from multiple threads — triggerAutoStop dispatches to DispatchQueue.global (a concurrent queue), and HybridAutoPlay.stopVoiceInput dispatches via a @MainActor Task. Two concurrent callers can both read isStopping as false, pass the guard, and each call continuation?.resume(returning: samples). Resuming a CheckedContinuation more than once is a fatal error in Swift that crashes the process. The Android counterpart avoids this by only resuming the continuation from the coroutine's finally block. A serial dispatch queue or a lock around stop() would prevent the double-resume.

packages/react-native-autoplay/ios/hybrid/HybridAutoPlay.swift
```diff
@@ -117,10 +119,53 @@ class HybridAutoPlay: HybridAutoPlaySpec {
...
+                interfaceController: interfaceController,
+                silenceThresholdMs: silenceThresholdMs ?? 1_500,
+                maxDurationMs: maxDurationMs ?? 10_000,
+                listeningText: listeningText ?? "Listening..."
+            )
+
+            HybridAutoPlay.voiceInputManager = nil
+
+            return try ArrayBuffer.copy(data: data)
+        }
```
SamuelBrucksch, line 161: Another comment from cursor:
> The Android startVoiceInput wraps the recording in try/finally to always run voiceInputManager = null and manager.dispose(), but the iOS version has no defer or equivalent. If manager.start() throws (e.g., converterUnavailable after the audio session was activated and the CPVoiceControlTemplate was already presented on CarPlay), HybridAutoPlay.voiceInputManager is never cleared and those partially-allocated resources — the active audio session and the on-screen voice template — are never cleaned up.

SamuelBrucksch (approved):

PR #74, opened 2026-04-21
fix(android): null-safe bitmap decode in fetchSync

Lucaaa03:
Hit this NPE during internal testing on an AAOS emulator:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'android.graphics.Bitmap android.graphics.Bitmap.copy(...)' on a null object reference
	at .../template/Parser.kt
```

`CloseableBitmap.underlyingBitmap` can be null and `.copy()` itself can throw (e.g. OOM on very large remote images, already-recycled bitmaps). Returning `null` instead of crashing lets the existing `CarIcon.ALERT` fallback kick in.

Follow-up to #67.


mefjuu:
PS. Are you planning to enable GH 'Issues' tab? BTW thanks for the library! I'm gonna try it soon.

g4rb4g3 (approved):

PR #75, opened 2026-04-24
Feature/save voice recordings

g4rb4g3:
This PR enhances the example app to allow it to save voice recordings.
Also fixes an issue that music playback would not pause while recording on iOS and brings back the missing list template example on CarPlay.


apps/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
```
g4rb4g3, line 4: fine on a none production example app

SamuelBrucksch (approved):

PR #88, opened 2026-06-02
feat: add isCarServiceRunning() to distinguish headless start sources

SamuelBrucksch:
## Summary

- Adds `isCarServiceRunning(): boolean` to the `AutoPlay` NitroModules spec, allowing apps to synchronously check whether the native car service (Android Auto's `CarAppService` or iOS's CarPlay scene) triggered the current headless execution.
- On Android, checks `AndroidAutoService.instance != null` — true as soon as the `CarAppService` is alive, before the session is fully established.
- On iOS, checks `SceneStore.isRootModuleConnected()` — true when the CarPlay scene is connected.

## Motivation

On Android, both Android Auto and other services (e.g. Firebase notifications) can start the app process headlessly. Until now there was no way to distinguish which source triggered the start. This made it impossible to skip heavy JS initialization (Redux store, middleware, subscriptions) during notification-only headless starts, causing the process to stay alive unnecessarily and waste battery.

`isCarServiceRunning()` is available immediately at startup before any component renders, allowing apps to gate their initialization in `index.js` using lazy `require()` calls.

## Usage

```js
import { HybridAutoPlay } from '@iternio/react-native-auto-play';

if (HybridAutoPlay.isCarServiceRunning()) {
  // AA/CP started the app — full initialization
} else {
  // Notification or other headless start — defer heavy imports
}
```

See `apps/example/index_with_headless.js` for a complete reference implementation covering all scenarios (notification-only, car headless, normal launch, late car connection).

## Test plan

- [x] Start Android Auto headlessly — verify `isCarServiceRunning()` returns `true`
- [x] Start app via notification without AA/CP — verify it returns `false`
- [x] Start app normally (tap icon) — verify it returns `false`
- [x] Connect CarPlay on iOS — verify it returns `true`
- [x] Verify `index_with_headless.js` pattern works for all four scenarios

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/react-native-autoplay/ios/hybrid/HybridAutoPlay.swift
+++ b/packages/react-native-autoplay/ios/hybrid/HybridAutoPlay.swift
@@ -83,6 +83,10 @@ class HybridAutoPlay: HybridAutoPlaySpec {
         return SceneStore.isRootModuleConnected()
     }
 
+    func isCarServiceRunning() throws -> Bool {
+        return SceneStore.isRootModuleConnected()
+    }
+
     func addSafeAreaInsetsListener(
         moduleName: String,
         callback: @escaping (SafeAreaInsets) -> Void
--- a/packages/react-native-autoplay/src/specs/AutoPlay.nitro.ts
+++ b/packages/react-native-autoplay/src/specs/AutoPlay.nitro.ts
@@ -131,11 +131,9 @@ export interface AutoPlay extends HybridObject<{ android: 'kotlin'; ios: 'swift'
   isConnected(): boolean;
 
   /**
-   * Check if the native CarAppService (Android Auto) is currently running.
-   * Use this to distinguish a headless execution triggered by Android Auto
+   * Check if the native AutoPlay is currently running.
+   * Use this to distinguish a headless execution triggered by AA / CP
    * from one triggered by other sources (e.g. notification updates).
-   * On iOS this always returns false.
-   * @namespace Android
    */
   isCarServiceRunning(): boolean;
 }
--- a/packages/react-native-autoplay/src/specs/AutoPlay.nitro.ts
+++ b/packages/react-native-autoplay/src/specs/AutoPlay.nitro.ts
@@ -129,4 +129,13 @@ export interface AutoPlay extends HybridObject<{ android: 'kotlin'; ios: 'swift'
    * @returns true if AutoPlay is connected, false otherwise.
    */
   isConnected(): boolean;
+
+  /**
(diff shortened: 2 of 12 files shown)


packages/react-native-autoplay/src/specs/AutoPlay.nitro.ts
```diff
@@ -129,4 +129,13 @@ export interface AutoPlay extends HybridObject<{ android: 'kotlin'; ios: 'swift'
    * @returns true if AutoPlay is connected, false otherwise.
    */
   isConnected(): boolean;
+
+  /**
+   * Check if the native CarAppService (Android Auto) is currently running.
+   * Use this to distinguish a headless execution triggered by Android Auto
+   * from one triggered by other sources (e.g. notification updates).
+   * On iOS this always returns false.
+   * @namespace Android
```
SamuelBrucksch:
```suggestion
```

packages/react-native-autoplay/src/specs/AutoPlay.nitro.ts
```diff
@@ -129,4 +129,13 @@ export interface AutoPlay extends HybridObject<{ android: 'kotlin'; ios: 'swift'
    * @returns true if AutoPlay is connected, false otherwise.
    */
   isConnected(): boolean;
+
+  /**
+   * Check if the native CarAppService (Android Auto) is currently running.
```
SamuelBrucksch:
```suggestion
   * Check if the native AutoPlay is currently running.
```

packages/react-native-autoplay/src/specs/AutoPlay.nitro.ts
```diff
@@ -129,4 +129,13 @@ export interface AutoPlay extends HybridObject<{ android: 'kotlin'; ios: 'swift'
...
+   * Use this to distinguish a headless execution triggered by Android Auto
```
SamuelBrucksch:
```suggestion
   * Use this to distinguish a headless execution triggered by AA / CP
```

packages/react-native-autoplay/ios/hybrid/HybridAutoPlay.swift
```diff
@@ -83,6 +83,10 @@ class HybridAutoPlay: HybridAutoPlaySpec {
         return SceneStore.isRootModuleConnected()
     }
 
+    func isCarServiceRunning() throws -> Bool {
```
SamuelBrucksch, line 86: Fixed

SamuelBrucksch:
@claude please review

g4rb4g3 (approved):

PR #90, opened 2026-06-11
Crucial ProGuard rule info

mefjuu:
Adding ProGuard rule is crucial to avoid startup crash in release mode when `def enableProguardInReleaseBuilds` is set to  `true`


g4rb4g3 (approved):
thank you very much :)

PR #93, opened 2026-06-22
feature/carplay panel

g4rb4g3:
This PR adds CarPlay CPMapPanel iOS 27 introduces. It is handled similar to what we do on Android Auto, giving the templates a mapConfig object puts the content in a panel instead of a template.

ListTemplate:
Main-VideoStream-2026-06-22-19-07-20

GridTemplate:
Main-VideoStream-2026-06-22-19-06-53

InformationTemplate:
Main-VideoStream-2026-06-22-19-07-50

MessageTemplate:
Main-VideoStream-2026-06-22-19-07-07

Also implements [navigationSession.optionsPanel](https://developer.apple.com/documentation/carplay/cpnavigationsession/optionspanel):
Main-VideoStream-2026-07-10-15-33-51
Main-VideoStream-2026-07-10-15-34-03
Main-VideoStream-2026-07-10-16-00-58


g4rb4g3:
@claude please review

g4rb4g3:
@claude 1. and 3. are fixexd. About popToRootTemplate, it removes all the panels that are not visible to the user anymore, this triggers the onPopped callback properly. All the other invisible callbacks have been called before when pushing a new panel on top of these. Only the top template is left on the stack so panelDidHide does all the cleanup properly. hidePanel() removes all the panels but only calls panelDidHide on the top most panel.

packages/react-native-autoplay/ios/scenes/AutoPlayInterfaceController.swift
```diff
@@ -239,14 +360,21 @@ class AutoPlayInterfaceController: NSObject, CPInterfaceControllerDelegate {
...
         )
 
-        if aTemplate is CPAlertTemplate {
-            try? RootModule.withTemplateStore { templateStore in
-                templateStore.removeTemplate(templateId: templateId)
-            }
+        /// this makes sure onPopped for templates is only invoked when they are gone forever
+        /// in case the template is just hidden by some other template it will not fire
+        /// since CPAlertTemplate is presented and nothing can be pushed on top of it there is no special handling required
+        guard !interfaceController.templates.contains(where: { $0.id == templateId }) else {
```
SamuelBrucksch, line 427: **Root swap evicts the MapTemplate from the store.**

Before this PR only `CPAlertTemplate`s were removed here; now any template missing from `interfaceController.templates` is removed. A root that gets replaced via `setRootTemplate` falls through this guard.

Scenario: `MapTemplate` (id `AutoPlayRoot`) is root → `listTemplate.setRootTemplate()` (e.g. sign-in) → `mapTemplate.setRootTemplate()` again. CarPlay fires `templateDidDisappear(map)`, `removeTemplate('AutoPlayRoot')` runs and `onPopped` fires. `HybridAutoPlay.setRootTemplate` begins with `templateStore.getTemplate`, so it and every later `HybridMapTemplate.*` call reject with `templateNotFound`. Since `MapTemplate` is a singleton that must not be constructed twice, the map is unusable until reconnect.

Suggest keeping the alert-only condition, or excluding the current/previous root from this cleanup.

packages/react-native-autoplay/ios/templates/InformationTemplate.swift
```diff
@@ -8,31 +8,69 @@
...
-            id: config.id
-        )
+        if #available(iOS 27.0, *) {
+            usesMapPanel = config.mapConfig != nil
+        }
+        else {
+            usesMapPanel = false
+        }
+
+        if !usesMapPanel {
```
SamuelBrucksch, line 65: **Crash on iOS < 27 with the new `PanelActionsIos` shape.**

The TS union permits `actions.ios: [TextButton, ImageButton]` whenever `mapConfig` is set. On iOS < 27 `usesMapPanel` is forced `false` and those same actions go to `Parser.parseInformationActions`, which does `CPTextButton(title: action.title!, …)` (Parser.swift:226). `NitroActionUtil.getTitle` returns `undefined` for an image button, so `title` is nil → force-unwrap crash at template construction, before anything is pushed.

Same path for `MessageTemplate` → `Parser.parseAlertActions` (`alertAction.title!`, Parser.swift:31).

Either filter/convert image actions in the fallback path, or gate the union on `Constants.isIos27OrGreater` on the JS side.

packages/react-native-autoplay/ios/hybrid/HybridAutoPlay.swift
```diff
@@ -192,23 +192,35 @@ class HybridAutoPlay: HybridAutoPlaySpec {
...
-                        animated: false
-                    )
-
-                    let _ = try await interfaceController.presentTemplate(
-                        carPlayTemplate,
-                        animated: animated
+                if #available(iOS 27.0, *), let panel = template.getPanel() as? CPMapPanel {
+                    /// since only the top most panel gets a panelDidHide call when pressing the close button
+                    /// we hide it globaly to avoid orphaned panels on the navigation stack
+                    panel.showsCloseButton = false
```
SamuelBrucksch, line 198: **Panels without iOS header actions have no way to be dismissed.**

`showsCloseButton = false` is forced for every panel, and `applyPanelHeaderActions(nil)` (AutoPlayMapPanelDelegate.swift:151) sets `backButton = nil` and empties both bar-button arrays. So `new ListTemplate({ title, sections, mapConfig: {} })` (the README example) leaves the driver with no control to leave the panel; only `popTemplate()` or `autoDismissMs` gets them out.

`MessageTemplate` is affected unconditionally: `MessageTemplate.ts` types `headerActions` as Android-only and converts `{ android: headerActions }`, so `getPanelHeaderActions()` is always nil on iOS. The JSDoc at MessageTemplate.ts:85 ("on iOS this template's own headerActions are applied to the root map template's nav bar") and README.md:926 ("the panel itself always shows its own non-customizable close/back control") both describe behaviour that doesn't exist.

Suggest either leaving `showsCloseButton` on, or synthesising a back button when no iOS header actions are supplied.

packages/react-native-autoplay/ios/scenes/AutoPlayInterfaceController.swift
```diff
@@ -49,99 +59,189 @@ class AutoPlayInterfaceController: NSObject, CPInterfaceControllerDelegate {
...
+    }
+
+    /// Pushes `panel` onto the current root template, which must already be a `CPMapTemplate`
+    /// (e.g. a `MapTemplate` previously set via `setRootTemplate`).
+    @available(iOS 27.0, *)
+    func pushPanel(
+        _ panel: CPMapPanel,
+        templateId: String
+    ) async throws {
+        guard let mapTemplate = rootTemplate as? CPMapTemplate else {
```
SamuelBrucksch, line 86: **`pushPanel` doesn't check that the map root is actually visible.**

Only `rootTemplate as? CPMapTemplate` is guarded. With stack `[root(map), ListTemplate B (no mapConfig)]`, pushing `InformationTemplate C` with `mapConfig: {}` attaches C to the covered root, `navigationStack` becomes `[root, B, C]`, nothing changes on screen and C's lifecycle callbacks never fire. The next `popTemplate()` hits the `.panel` case in `popTopEntry` and pops the invisible C instead of the visible B, so JS and the screen are out of sync.

Suggest throwing (or falling back to a regular push) when `interfaceController.topTemplate !== rootTemplate`.

packages/react-native-autoplay/ios/templates/ListTemplate.swift
```diff
@@ -8,44 +8,98 @@
...
         }
 
+        if #available(iOS 27.0, *), let mapPanel = mapPanel as? CPMapPanel {
+            mapPanel.sections = Parser.parseMapPanelSections(
+                sections: sections,
+                updateSection: self.updateSection(section:sectionIndex:),
+                traitCollection: traitCollection
+            )
+
+            return
```
SamuelBrucksch, line 97: **`setHeaderActions()` is a silent no-op while this template is shown as a panel.**

`HybridAutoPlay.setTemplateHeaderActions` does `barButtons = …; invalidate()`, but this `_invalidate` returns right after updating `sections` and `applyPanelHeaderActions` is only invoked from `panelDidShow` / `handlePanelPopped`. So the promise resolves, `barButtons` is updated, and the map's nav bar keeps the old buttons until the panel is re-shown.

Same pattern in GridTemplate.swift:88 and InformationTemplate.swift:112. The panel branch should re-apply header actions (and map buttons) when this template is the visible panel.

packages/react-native-autoplay/ios/templates/MapTemplate.swift
```diff
@@ -176,15 +117,35 @@ class MapTemplate: AutoPlayHeaderProviding,
...
-        setBarButtons(template: template, barButtons: barButtons)
+        // A panel owns the bar buttons while it exists; keep barButtons up to date but don't apply it.
+        let hasPanel: Bool = {
+            if #available(iOS 27.0, *) {
+                return SceneStore.getRootScene()?.interfaceController?.hasPanel ?? false
+            }
+            return false
+        }()
+
+        if !hasPanel {
```
SamuelBrucksch, line 136: **Day/night switch leaves panel bar/map buttons in the old colours.**

`_invalidate` skips `setBarButtons` and `parseMapButtons` whenever `hasPanel`, and `Parser.parseMapButtons` / `parseHeaderActions` now render glyphs with `noImageAsset: true` (colour picked once at parse time, not a light/dark `UIImageAsset`). `applyPanelHeaderActions` / `applyPanelMapButtons` only run from `panelDidShow` and `handlePanelPopped`, and none of the panel-backed templates override `traitCollectionDidChange`.

Result on iOS 27: a `ThemedColor` glyph map button stays black after the head unit switches to night mode until the panel is popped. Pre-PR the `UIImageAsset` variant switched automatically.

The trait-change path needs to re-apply the visible panel's buttons (or `MapTemplate` should own the buttons and consult a panel override, so one writer handles both cases).

packages/react-native-autoplay/ios/scenes/AutoPlayInterfaceController.swift
```diff
@@ -49,99 +59,189 @@ class AutoPlayInterfaceController: NSObject, CPInterfaceControllerDelegate {
...
         _ rootTemplate: CPTemplate,
         animated: Bool
     ) async throws -> Bool {
-        return try await interfaceController.setRootTemplate(
+        let result = try await interfaceController.setRootTemplate(
             rootTemplate,
             animated: animated
         )
+
+        navigationStack = [.template(id: rootTemplate.id)]
```
SamuelBrucksch, line 142: **`setRootTemplate` drops `.panel` entries without tearing them down.**

Regular templates are cleaned up via `templateDidDisappear`, but panels have no equivalent CarPlay callback, and `panelDidHide`'s `stillTracked` guard is already false after the stack reset, so `handlePanelPopped` is skipped too.

Net effect on a root swap with panels open: the `AutoPlayTemplate`s stay in `TemplateStore` (holding `CPMapPanel`, delegate and all JS callbacks), their listeners stay in `HybridAutoPlay`, and `onPopped` never reaches JS.

Suggest collecting the `.panel` ids before reassigning and running them through `templateStore.removeTemplates` + `HybridAutoPlay.removeListeners`.

packages/react-native-autoplay/ios/templates/MapTemplate.swift
```diff
@@ -846,13 +812,60 @@ class MapTemplate: AutoPlayHeaderProviding,
...
+        guard let navigationSession else { return }
+        guard let optionsPanel else {
+            navigationSession.optionsPanel = nil
+            return
+        }
+        guard let traitCollection = SceneStore.getRootTraitCollection() else { return }
+
+        let sections = Parser.parseOptionsPanelSections(
+            sections: optionsPanel.sections,
+            updateSection: updateOptionsPanelSection(section:sectionIndex:),
```
SamuelBrucksch: **Retain cycle via `updateSection`.**

`updateOptionsPanelSection(section:sectionIndex:)` is an unapplied method reference that captures `self` strongly; Parser stores it in every options-panel `CPListItem.handler`. That gives `MapTemplate → navigationSession → optionsPanel → items → handler → MapTemplate`.

Only `stopNavigation()` (which nils the session) breaks it, and `templateApplicationScene(_:didDisconnectInterfaceController:)` calls `disconnect()` without it, so a MapTemplate mid-navigation leaks along with its session and all captured JS `onPress` callbacks.

```swift
updateSection: { [weak self] section, index in
  self?.updateOptionsPanelSection(section: section, sectionIndex: index)
}
```

packages/react-native-autoplay/src/templates/InformationTemplate.ts
```diff
@@ -47,35 +51,54 @@ export type InformationTemplateConfig = Omit<
...
-        ]
-      | [TextButton<InformationTemplate>, TextButton<InformationTemplate>]
-      | [TextButton<InformationTemplate>];
-  };
 };
 
+export type InformationTemplateConfig = InformationTemplateBaseConfig &
+  (
+    | {
+        mapConfig?: undefined;
```
SamuelBrucksch, line 67: **Breaking change for existing Android `mapConfig` passthrough configs.**

`InformationTemplateConfig` / `MessageTemplateConfig` were flat types with `mapConfig?: BaseMapTemplateConfig` and up to three iOS text actions. As a union discriminated on `mapConfig`, a config with `mapConfig: props?.mapConfig` (type `X | undefined`) and 2–3 iOS text actions no longer compiles (I checked with a scratch `tsc` run: TS2345). The example app had to be split into two constructor calls with an explicit `mapConfig: undefined` for the same reason.

At runtime with `mapConfig` set, `Parser.parsePanelButtonConfiguration` only reads `actions[0]` as text and `actions[1]` as an icon, so a third action is dropped and a second text action loses its title.

Worth calling out in the changelog as breaking, or keeping the flat type and validating the panel case at runtime.

packages/react-native-autoplay/src/templates/ListTemplate.ts
```diff
@@ -22,58 +21,15 @@ const HybridListTemplate = NitroModules.createHybridObject<NitroListTemplate>('L
...
-      title: string;
-      items: Array<RadioRow<T>>;
-    };
-
-export type SingleSection<T> = {
-  [K in MultiSection<T> as K['type']]: Omit<K, 'title' | 'detailedText'>;
-}[MultiSection<T>['type']];
-
-export type Section<T> = Array<MultiSection<T>> | SingleSection<T>;
+export type {
```
SamuelBrucksch, line 24: **New public types aren't exported from the entry point.**

`WaypointRow` / `WaypointCoordinate` are missing from this list, and nothing re-exports `OptionsPanelConfig`, `OptionsPanelSection`, `ChargerOutlet`, `ChargerLocation`, `ChargingConnector` from `utils/NitroOptionsPanel.ts` (it's only imported by `MapTemplate.ts`). The README documents `OptionsPanelConfig` and `ChargerLocation` by name, but `import type { WaypointRow, OptionsPanelConfig } from '@iternio/react-native-auto-play'` fails with TS2305.

The example currently works around it with `import type { WaypointRow } from '@iternio/react-native-auto-play/lib/utils/NitroSection'` (apps/example/src/templates/AutoListTemplate.ts:12). `lib/` is gitignored build output with no `exports` map, so `yarn typecheck:example` only passes after `yarn build:auto-play` has run.

SamuelBrucksch (commented):
Automated review (high effort) of the iOS panel / options-panel changes. The library typecheck, lint and full build (including nitrogen regeneration) pass cleanly on the PR head; `yarn lint:example` passes with 4 `noFloatingPromises` warnings on the new `popTemplate()` / `popToRootTemplate()` calls in `AutoListTemplate.ts` (lines 50, 57, 139, 147 — the `AutoMessageTemplate.ts` changes use `void …` correctly).

10 confirmed correctness findings are attached inline. Lower-priority items not commented inline:

- **Plausible:** the cover/reveal simulation in `templateWillAppear` assumes the covering entry is the last `navigationStack` element, but presented `CPAlertTemplate`s are never in the stack, so dismissing a plain `MessageTemplate` over a panel gives the panel no `onWillAppear`/`onDidAppear` pair (or the wrong panel with two open).
- **Plausible:** `updateOptionsPanel` → `applyOptionsPanel` mutates `navigationSession.optionsPanel` and renders glyphs on the calling (JS) thread with no `@MainActor` hop, unlike `setTemplateMapButtons`. Same as the pre-existing `startNavigation` pattern, so not a regression on its own.
- Options-panel `list` sections default a missing `title` to `''` rather than `undefined`, producing an empty section header on CarPlay (grid/charger sections keep `nil`).
- Every toggle/radio tap in the options panel rebuilds the whole `CPMapPanel` (all sections re-parsed, all glyphs re-rendered, new panel assigned); `ListTemplate` already shows in-place `mapPanel.sections = …` is available.
- The `mapPanel` / `mapPanelDelegate` / `getPanel*` plumbing is copy-pasted into `ListTemplate`, `GridTemplate`, `InformationTemplate` and `MessageTemplate` (~80 lines) and has already drifted (eager vs lazy construction; `MessageTemplate` returns `config.headerActions` where the others return `barButtons`). `AutoPlayTemplate` already declares the overridable defaults, so storage + delegate wiring could live there.
- `popToRootTemplate` re-fetches itself via `RootModule.withInterfaceController` although `self.rootTemplate` and the kept `.panel` entry's `mapTemplate` are already in scope.

packages/react-native-autoplay/ios/scenes/AutoPlayInterfaceController.swift
g4rb4g3: Setting a new root template is not a common flow — it tears down the React root view, which effectively stops the whole CarPlay app on that surface. So onPopped firing for the old root (including MapTemplate) here is correct: it's gone for good and shouldn't be reused. Bringing CarPlay back to a map afterwards means constructing a new MapTemplate, not reusing the old instance.

packages/react-native-autoplay/ios/hybrid/HybridAutoPlay.swift
g4rb4g3: updating the readme to make this clear, the close button has to be disabled as it would break the lifecycle tracking otherwise.

packages/react-native-autoplay/src/templates/InformationTemplate.ts
g4rb4g3: updating docs and add breaking warning to release

SamuelBrucksch:
@claude please review this again. check the comments in this PR also and consider the previous reviews if possible.

(mainly a test if the new fable reviewer works)

g4rb4g3:
@claude check the latest commit and see if it fixes the issues you mentioned on the last review.

SamuelBrucksch (approved):
Good Job Manuel and Claude.

PR #94, opened 2026-06-24
fix: telemetry observers executing on main thread

g4rb4g3:
This PR makes sure the callbacks for Android Auto and Android Automotive telemetry data are run on the proper AndroidTelemetryThread thread instead of the main thread.


g4rb4g3:
@claude please review

PR #97, opened 2026-07-01
Feature/voice input sounds

g4rb4g3:
This PR adds support for sounds to be played back just before starting voice input and after it has ended.
Also makes `HybridVoice.startVoiceInput` reject in case the user cancels.


g4rb4g3:
@claude please review

SamuelBrucksch (approved):

PR #99, opened 2026-07-14
fix: fill Android Auto surface after host resizes it

SamuelBrucksch:
## Problem

On some head units (reported: Nothing Phone / Pixel on XPeng, where AA runs windowed) the app renders at a smaller size in the top-left corner, with the rest of the screen black. The host signals size changes only via `onSurfaceDestroyed` → `onSurfaceAvailable` with new dimensions, but `VirtualRenderer` only recreated the `VirtualDisplay` — view layout params, scale and Fabric measure specs stayed frozen at first-surface values.

## Fix

- `VirtualRenderer` recomputes all size-derived state on every `onSurfaceAvailable`: VirtualDisplay, density/scale (now derived from `surfaceContainer.dpi` instead of the CarContext), view layout params and Fabric layout specs (`updateRootLayoutSpecs`). The react surface stays alive, so JS state survives the resize — no app restart.
- Stale visible/stable-area state is reset on surface destroy so insets aren't computed from old rects.
- New Android-only nitro module `AndroidWindowInformation` emits window dimension changes. Root components get their `window` prop updated automatically via an internal wrapper — no app-side changes needed. The listener is also exported for logic outside root components.

iOS/CarPlay is unaffected: no native or generated iOS changes; the wrapper is a pass-through there.

## Testing

Verified on device by simulating host resizes (80% ↔ full surface): content fills the new size, JS state (sign-in, counters) survives, `props.window` updates live.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SamuelBrucksch:
@claude please review

g4rb4g3 (approved):

PR #104, opened 2026-07-16
Permit non-navigation Android Auto app types

jzarnett:
This PR introduces the option to declare types other than NAVIGATION for Android Auto.

Greetings. I've been working on adding Android Auto support to an application that has
the concept of a schedule. That is, the user has a list of places they need to go and
jobs to do when they get there, that represent their calendar/assignments for the day.
In Apple CarPlay, this is represented as a "Driving Task" app, and we've sought and
already received approval from Apple to publish the app in the app store with that
deisgnation. For Android Auto, though, this library always defaults the declared
type to `NAVIGATION`, which is intended for something like Waze: draw a map, turn by
turn directions, etc. But the app for which I want to use this is not really a maps
application, nor should it be -- it has a list of jobs to be done and a navigate
button that sends the user over to Google Maps (or Waze). If the app I'm working on
is declared as `NAVIGATION` it is rejected in the Google Play store as the testers
are unable to verify how to activate turn by turn directions, et cetera. 

As a workaround in my app I've applied a patch to the library that turns the type
of app into `POI` (which we think is the best match for this use case, even if it
is not a 1:1 match for Apple's Driving Task deisgnation). But then the thought came
to me that I should look to upstream this capability rather than rely on a patch of
the library. I imagined others would benefit from this as well.

According to the docs as of 2026-07-16, there are more types than just navigation
or POI, including messaging, calling, Internet of Things, and weather. 

https://developer.android.com/training/cars

This is, of course, my first (potential) contribution so feedback is welcome.

Acknowledgement: I did get Claude Code to write the changes here.

### DIFF
--- a/packages/react-native-autoplay/README.md
+++ b/packages/react-native-autoplay/README.md
@@ -206,6 +206,12 @@ In case you have ProGuard enabled (`def enableProguardInReleaseBuilds = true` in
 ### Android Auto Customization
 You can customize certain behaviors of the library on Android Auto by setting properties in your app's `android/gradle.properties` file.
 
+-   **App Category**: Declare which Android Auto / Automotive app type the `CarAppService` advertises — the Android counterpart to choosing your CarPlay scene type in `Info.plist`.
+    ```properties
+    ReactNativeAutoPlay_androidAutoAppCategory=navigation
+    ```
+    The value maps to `androidx.car.app.category.<UPPERCASE>` (e.g. `navigation`, `poi`, `parking`, `charging`) — see the [supported app categories](https://developer.android.com/training/cars). The default `navigation` keeps the existing behavior. Any other value selects a lean manifest that declares the chosen category and omits the navigation-only permissions (`NAVIGATION_TEMPLATES`, `MAP_TEMPLATES`, `ACCESS_SURFACE`), the `FEATURE_CLUSTER` category, and the `NAVIGATE`/`geo` intent filters — a non-navigation, no-map template app needs none of these, and declaring them gets the app reviewed against the navigation-app bar on the Play Store.
+
 -   **Telemetry Update Interval**: Control how often telemetry data is updated.
     ```properties
     ReactNativeAutoPlay_androidTelemetryUpdateInterval=4000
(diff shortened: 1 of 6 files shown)


SamuelBrucksch:
@claude please review

jzarnett:
Thanks for the review! Addressed in 51b5535.

**1. No input validation on `androidAutoAppCategory`** — Fixed. `build.gradle` now validates the value against the supported set (`navigation`, `poi`, `parking`, `charging`, `iot`, `messaging`, `calling`, `weather`) and fails the build with a clear message on anything else. The check lives inside `defaultConfig` (after `compileSdk` is set) so it surfaces as a single clean error rather than an added "compileSdk not specified" one. Also normalized casing with `Locale.ROOT` so tokens containing `i` aren't mangled under a Turkish-locale build.

**2. `NotificationCompat.CATEGORY_NAVIGATION` hardcoded** — Fixed. Added an `IS_NAVIGATION_APP` `BuildConfig` field derived from the same Gradle property; the foreground-service notification now uses `CATEGORY_NAVIGATION` for navigation apps and `CATEGORY_SERVICE` otherwise.

**3. `HybridAndroidAutomotive` unconditionally uses `APP_FOCUS_TYPE_NAVIGATION`** — Documented, as suggested. The Car API has no non-navigation app-focus type, so the correct guidance is not to call these helpers for non-navigation apps; the README now states the app-focus helpers request/observe navigation focus specifically and shouldn't be used with a non-navigation category. Happy to file a follow-up if we want stricter guarding.

**4. README lists only a subset of categories** — Fixed. The README now lists all supported values and notes that an unrecognized value fails the build.

Verified against a real AGP build: invalid category fails cleanly; `navigation` produces the navigation manifest with `IS_NAVIGATION_APP=true`; `poi` produces the lean manifest with `IS_NAVIGATION_APP=false`; the library Kotlin compiles.

g4rb4g3 (approved):
Nice stuff, thank you for your effort.
I'm curios on what apps use this lib so if you don't mind you can drop your app name here. :)

jzarnett:
> I'm curios on what apps use this lib so if you don't mind you can drop your app name here. :)

Thanks for your review. I don't mind saying at all. This is for Jobber ( www.getjobber.com ). This is a field service software app. For example, the customer for the app is a plumber: they have a schedule of their clients to visit today, so we want to be able to show that on the screen of their work truck when they connect to Android Auto (or Carplay). Then they can start navigation from the app.

PR #105, opened 2026-07-17
fix(react-native-autoplay): route CarPlay maneuver estimates correctly

SamuelBrucksch:
## Summary

Fixes two CarPlay travel-estimate bugs in `MapTemplate.swift`:

- **Estimate misrouted onto lane secondary.** The update indexed the unfiltered `navigationSession.upcomingManeuvers` with an index (`maneuverIndex`) computed against the *filtered* `sessionManeuvers` (non-secondary) list. When lane guidance was present, `-lanes` secondary maneuvers shifted the indices, so the estimate landed on a secondary maneuver and the real maneuver stayed stuck on its stale value. Now updates `sessionManeuvers[maneuverIndex]`.
- **Stale distance on maneuver promotion.** When a maneuver is promoted to current (previous turn completes, next moves to front) it is reused and keeps the `initialTravelEstimates` baked in at creation until the next periodic update. We now push the freshest estimate onto the current maneuver immediately, so its distance isn't briefly stuck on the previous turn's value.

## Test plan

- [x] Test in the car and make sure the instruction has proper distance in HUD when it updates

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/react-native-autoplay/ios/templates/MapTemplate.swift
+++ b/packages/react-native-autoplay/ios/templates/MapTemplate.swift
@@ -729,11 +729,16 @@ class MapTemplate: AutoPlayHeaderProviding,
                 sessionManeuvers
                 .firstIndex(where: { $0.id == nitroManeuver.id })
             {
+                // `maneuverIndex` indexes `sessionManeuvers` (the filtered,
+                // non-secondary list), so update that same maneuver. Indexing the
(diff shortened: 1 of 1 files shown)


SamuelBrucksch:
@claude please review

g4rb4g3 (approved):

PR #106, opened 2026-07-28
Feature/voice input encoding

g4rb4g3:
This PR makes sure voice-input chunks are ~100ms each and adds ALAW and MULAW encoding support.


g4rb4g3:
@claude please review

PR #107, opened 2026-07-30
fix: avoid releasing display too early to prevent freezes on some devices

g4rb4g3:
## Problem
On some Samsung devices (confirmed: Galaxy S25, Android 16), the
Android Auto map and RN components could freeze indefinitely. Root cause:
Samsung's BBA power-management logic suspends the main thread's
Choreographer if it sees the app's window disappear, even briefly.
VirtualRenderer used to release the outgoing VirtualDisplay/Presentation
in onSurfaceDestroyed -- which could leave no AA window registered
while a full-screen template was shown.

Once suspended, everything riding on Choreographer stops: Mapbox's
ValueAnimator-driven camera/puck, RN's setInterval/setTimeout, and RN
view layout/draw for components. Nothing crashes -- it just goes
silent until something resets Samsung's classifier (e.g. waking the
phone screen).

https://github.com/user-attachments/assets/06a106f4-e3e5-4c3c-bafc-3d2eccfa059e

## Fix
Stop releasing displays in onSurfaceDestroyed (also not reliably
invoked on all hosts). Instead, stash the outgoing display and only
release it once the replacement Presentation has confirmed a drawn
frame, so the AA window is never absent even momentarily.

https://github.com/user-attachments/assets/b9b09a0d-08a1-4d8b-81f5-68640d85309e

### DIFF
--- a/packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/VirtualRenderer.kt
+++ b/packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/VirtualRenderer.kt
@@ -41,6 +41,7 @@ class VirtualRenderer(
     private val isCluster: Boolean = false
 ) {
     private var virtualDisplay: VirtualDisplay? = null
+    private val pendingDisplays = mutableListOf<VirtualDisplay>()
 
     private var reactSurfaceImpl: ReactSurfaceImpl? = null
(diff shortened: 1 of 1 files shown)


g4rb4g3:
@claude please review

SamuelBrucksch (approved):

g4rb4g3:
Confirmed to fix the issue users reported on ABRP: https://abrp.featurebase.app/de/p/android-auto-map-freezes-when-navigating-a-trip-plan-with/comment/6a748275a4c7cd190a8d76f1

PR #109, opened 2026-08-25
feat: optional background fragment for remote images

forki:
## What

- `parseRemoteImage` (Android + iOS) now recognizes an optional `#bg=RRGGBB` fragment on the image URI and renders the fetched image centered on a rounded, colored background before tinting/caching.
- `NitroImage`'s asset conversion now handles local `file://` and `data:` URIs directly instead of routing them through `Image.resolveAssetSource`, which only resolves bundled `require(...)` assets.

## Why

`#bg=RRGGBB`: useful for logos/icons designed for a light background when displayed on a dark host surface (or vice versa) — avoids having to pre-bake a background into every asset variant, and keeps the choice at the call site (`{ uri: 'https://.../logo.png#bg=1E1E1E' }`).

`file://`/`data:` bypass: `Image.resolveAssetSource` silently returns such sources unchanged with `width`/`height` left `undefined`, which breaks NitroModules' strict typing further down the pipeline. Handling them explicitly makes locally-generated/cached images (e.g. from a native file cache) usable the same way as `require(...)` assets.

## Testing

Verified manually against a downstream app consuming this library. No test suite changes — the touched files (`Parser.kt`, `Parser.swift`, `NitroImage.ts`) don't currently have unit test coverage in this repo.


g4rb4g3:
@claude please review

packages/react-native-autoplay/src/utils/NitroImage.ts
```diff
@@ -119,6 +119,26 @@ function convert(image?: AutoImage): NitroImage | undefined {
...
+    return {
+      height: 0,
+      scale: 1,
+      uri: image.image.uri,
+      width: 0,
+      packager_asset: false,
+      color: NitroColorUtil.convert(image.color),
+    };
+  }
+
```
g4rb4g3: What is the reason for this?
Image.resolveAssetSource and the defaults below should handle this already, right?

g4rb4g3 (changes requested):
Thanks for the PR @forki, could you please check the comments? Thank you!

The corner radius might be fine since AA and CP behave different. The other ones are minor things that are easy to fix probably.

forki:
Rebased onto current master (0.6.0) and went through the review points.

**You were right about the `file://`/`data:` guard — I removed it.** I checked the RN 0.86 sources: `resolveAssetSource` returns any non-number source unchanged (`Libraries/Image/resolveAssetSource.js:112`), so a `{ uri }` object already falls through with `uri` set, and the defaults below fill in the rest. The only difference the guard made was `scale: 1` instead of `scale: 0`, and `RCTConvert UIImage:` skips the rescale entirely when scale is 0 (`React/Base/RCTConvert.mm:1358-1366`), while the Android path ignores scale completely. So it produced identical output and was dead weight. The PR is now just the background fragment; the title reflects that.

**Android fragment stripping now matches iOS** — the fragment is only stripped when a `#bg=` was actually parsed, instead of unconditionally.

**Corner radius is now the same proportion on both platforms** (`size * 0.125`, i.e. 32 on a 256 canvas) instead of a hardcoded 20 on iOS. You mentioned the two platforms may legitimately differ, but since 20pt on a 256pt iOS canvas and 32px on a 256px Android canvas are different proportions (7.8% vs 12.5%), matching them removes the difference the review flagged without guessing at platform behaviour. Say the word if you would rather keep iOS at 20.

PR #110, opened 2026-08-25
feat: head-unit navigation (Android Auto + CarPlay)

forki:
## What

Adds `AutoPlayNavigation.navigate(lat, long, label)` as a native module, for apps that need to start turn-by-turn navigation to a destination from JS while running on the head unit.

## Why

`Linking.openURL` doesn't work in either headless context:
- **Android Auto**: the app runs as a `CarAppService` (service context, no Activity), and `Linking.openURL` needs `startActivity()` internally.
- **CarPlay**: the sandbox blocks `UIApplication.shared.open()` with `FBSOpenApplicationErrorDomain Code=3 "Request is not trusted"`.

Android uses `CarContext.startCarApp(Intent(ACTION_NAVIGATE, geo:...))`. iOS uses `MKMapItem.openInMaps(fromScene:)` (Apple's recommended API for CarPlay navigation, iOS 13.2+), guarded with a `#if TARGET_OS_SIMULATOR` early-reject — opening Maps as a CarPlay scene on the Simulator leaves an unrecoverable blank screen there.

## Usage

```ts
import { AutoPlayNavigation } from '@iternio/react-native-auto-play';

if (AutoPlayNavigation) {
  await AutoPlayNavigation.navigate(51.49, 11.93, 'Destination label');
}
```

## Note on overlap with #109 and a possible follow-up PR

This PR and #109 (remote image background) are independent. I also have a `PointOfInterestModule` (`PlaceListMapTemplate` / `CPPointOfInterestTemplate` wrapper) coming as a separate PR that touches the same `ReactNativeAutoPlayPackage.kt` registration switch as this one — happy to rebase whichever lands second.

## Testing

Verified manually against a downstream app (Android Auto + CarPlay, incl. the CarPlay Simulator guard). No existing unit test coverage for native module registration in this repo.


g4rb4g3 (changes requested):
Nice, this allows to call some navigation app running on AA/CP, right?
As we use nitro-modules this should not be a native module but rather be added to AutoPlay.nitro.ts or a new spec.

forki:
Rebased onto current master (0.6.0) and converted to nitro-modules as requested.

`AutoPlayNavigationModule` (the Kotlin `ReactContextBaseJavaModule` and the ObjC `RCTBridgeModule`) is gone, along with its registration in `ReactNativeAutoPlayPackage`. The capability is now `navigate(latitude, longitude, label)` on the existing `AutoPlay` spec — I went with adding it to `AutoPlay.nitro.ts` rather than a new spec since that avoids a new `nitro.json` autolinking entry and the call belongs to the same object that already owns template/session control. Happy to split it into its own spec if you prefer the separation.

- **Android** (`HybridAutoPlay.kt`): `AndroidAutoSession.getRootContext()` → `CarContext.startCarApp(ACTION_NAVIGATE, geo:…)`.
- **iOS** (`HybridAutoPlay.swift`): `MKMapItem.openInMaps(launchOptions:fromScene:)` with the `CPTemplateApplicationScene` taken from `SceneStore.getRootScene()?.window?.windowScene`, instead of walking `UIApplication.shared.connectedScenes`.
- Nitrogen output regenerated (`yarn specs`); `tsc --noEmit` and `biome check src` are clean.

Note I cannot compile Kotlin or Swift here (no Android SDK / Xcode on this machine), so the native sides are review-only — the Kotlin and Swift both mirror the structure of the neighbouring methods, but a build on your side is the real check.

PR #111, opened 2026-08-25
feat: Android Auto loading screen with spinner instead of static icon

forki:
## What

Replaces the root-session `MessageTemplate` (app icon on white background) shown while waiting for JS to render its first screen with a `PaneTemplate(setLoading(true))`, which the host renders with a loading spinner plus a title.

## Why

The static-icon `MessageTemplate` reads as a hang to some users — there's no visual indication anything is happening. `PaneTemplate` with `setLoading(true)` gets a host-drawn spinner. Title defaults to `"Loading <AppName>…"`, overridable via a host string resource named `AutoPlayLoadingLabel` (same convention `getApplicationLabel` already uses for `AutoPlayClusterSplashScreenLabel`). Falls back to the previous `MessageTemplate` if the host ever rejects `PaneTemplate`.

Cluster sessions are unaffected — they always render a `MapTemplate` and never hit this template's fallback branch.

## Testing

Verified manually on Android Auto (root session cold start). No existing unit test coverage for `AndroidAutoSession`/`AppInfo` in this repo.


g4rb4g3 (changes requested):
While I see the reason for someone to want this I think it does not look great... 😬 
Can we have a flag to opt-in on the PaneTemplate and keep the default MessageTemplate instead? You can do it in a similar way then androidTelemetryUpdateInterval for example.

forki:
Rebased onto current master (0.6.0) and made it opt-in as requested.

The `MessageTemplate` is the default again — nothing changes for existing users. The `PaneTemplate` with the spinner now requires `ReactNativeAutoPlay_androidLoadingTemplate=true` in `android/gradle.properties`, wired up the same way as `androidTelemetryUpdateInterval` (Gradle property → `BuildConfig.LOADING_TEMPLATE`). Documented in the README under *Android Auto Customization*.

The title stays overridable via the `AutoPlayLoadingLabel` string resource, and if the host rejects the `PaneTemplate` it still falls back to the `MessageTemplate`.

PR #112, opened 2026-08-25
fix: voice-navigation intents lost on cold start / startup race

forki:
## What

Fixes two related ways an OS voice-navigation intent (`CarContext.ACTION_NAVIGATE`) can silently fail to reach JS.

## Why

1. **Cold start**: `onCreateScreen` ignored the launch intent entirely — only `onNewIntent` (warm path) parsed it. If the OS starts the app cold in response to a voice command ("navigate to the nearest coffee shop"), the query was dropped on the floor. Fixed by capturing the launch intent and replaying it once the screen reaches `DIDAPPEAR`, by which point JS has had a chance to register its listener.
2. **Startup race** (independent of #1): `HybridAutoPlay.emitVoiceInput` was fire-and-forget. If the OS delivers the intent before JS ever calls `addListenerVoiceInput` (slow `onConnect`/`trySetRoot`), the event is lost forever — with identical app code, this makes voice work in one build and not another purely depending on how fast the JS bundle happens to start. Fixed by buffering the most recent emit when there's no listener yet, replayed to the first listener that registers within a 30s TTL.

Also refactored `onNewIntent`'s inline parsing into a reusable `handleVoiceIntent(intent): Boolean` so the warm path and the buffered cold-start replay share one implementation instead of duplicating the geo-URI parsing.

## Testing

Verified manually on a real Android Auto head unit across both cold-start and warm-path voice commands, with artificial JS-side delay to reproduce the startup race. No existing unit test coverage for `AndroidAutoSession`/`HybridAutoPlay` in this repo.


g4rb4g3 (changes requested):
Good catch! 👍 

Imho we do not need this entire `pendingVoiceIntent` handling in `AndroidAutoSession`. We can simply call `handleVoiceIntent` from `onCreateScreen`, in case it is a valid voice input intent `HybridAutoPlay.emitVoiceInput` takes care of emitting it directly in case a listener is registered or stores it for later registered listeners.
Do you agree or did I miss something?

Besides that, please stick to short and precise comments wherever they provide a meaningful value. No need to comment obvious things. 🙂 Thanks!

forki:
Rebased onto current master (0.6.0) and addressed the feedback.

**`pendingVoiceIntent` handling is gone** — you are right, it was redundant. `handleVoiceIntent` is now called directly from `onCreateScreen`; `HybridAutoPlay.emitVoiceInput` buffers the event when no listener is registered yet and replays it to the first listener that shows up, so no replay timer and no lifecycle dance are needed.

One detail I kept: the call in `onCreateScreen` is still guarded by `clusterId == null`. `AndroidAutoSession` is instantiated for the cluster display as well, and a cluster session would emit the same launch intent a second time — `emitVoiceInput` is a global emit, so JS would see the voice event twice. Happy to drop the guard if you think a cluster session never sees that intent.

**Comments trimmed** — the long rationale blocks in `AndroidAutoSession.kt` and `HybridAutoPlay.kt` are cut down to what is not obvious from the code.

PR #113, opened 2026-08-25
feat: PointOfInterestTemplate (PlaceListMapTemplate / CPPointOfInterestTemplate)

forki:
## What

Adds a `PointOfInterestTemplate` JS wrapper around a pinned-places list — `PlaceListMapTemplate` on Android Auto, `CPPointOfInterestTemplate` on CarPlay — for "find nearby X" style flows, backed by a new `PointOfInterestModule` native module on both platforms.

```ts
import { PointOfInterestTemplate } from '@iternio/react-native-auto-play';

const template = new PointOfInterestTemplate({
  title: 'Nearby places',
  items: [{ id: '1', title: 'Place A', lat: 51.49, lng: 11.93, distanceMeters: 420 }],
  onSelectItem: (id) => { /* ... */ },
});
template.push();
```

## Notable behavior baked into the initial implementation

Since this is a new template with no prior version, these aren't "fixes" against previous behavior, but worth calling out for review:

- `PlaceListMapTemplate` requires a `DistanceSpan` on every non-browsable row, or the host shows a meaningless placeholder distance instead of the real one. Implemented as a spanned leading character on `line1`.
- `Distance.create()` renders the given value as-is in the given unit — it does not auto-convert. Passing raw meters unconditionally with `UNIT_METERS` would show `"134363.5 m"` instead of `"134.4 km"` for a 134 km route; switches to kilometers above 1000m.
- `setCurrentLocationEnabled(true)` throws a `SecurityException` that crashes the whole car app session if the location permission was denied at the OS level — gated behind an actual permission check.

## Open design question: pin rendering

`PinBitmapRenderer` (Android) and the equivalent function in `PointOfInterestModule.m` (iOS) are an opinionated default — a 3-state status color with an `"available/total"` label — not a generic pin API. Consumers with a different marker model will want their own renderer with the same shape. Happy to discuss extracting this behind a customization hook (e.g. a JS-side render callback, or letting callers pass a pre-rendered image URI per item) if that's preferred over shipping an opinionated default.

## Note on overlap with #110

This PR's `ReactNativeAutoPlayPackage.kt` change and #110's are independent additions to the same registration `when`/`hashMapOf` — happy to rebase whichever lands second.

## Testing

Verified manually on a real Android Auto head unit and CarPlay (device + simulator, including the empty-POI-list and location-permission-denied edge cases). No existing unit test coverage for template/module registration in this repo.


g4rb4g3 (changes requested):
We are not using native modules on this project but rather nitro-modules.
Please check the other templates on their implementation and align this one. Thank you!
Besides that:
- make sure to implement all the on press listeners correctly, I see some `CarToast.makeText` for example that looks like debug code to me.
- Do not use `RCTDeviceEventEmitter` as nitro-modules gives some nice callbacks.
- I see some hardcoded colors, can we make them configurable?

forki:
Rebased onto current master (0.6.0) and reworked along all of your points.

**No more native modules — it is a real template now.** `PointOfInterestModule.kt`, `ReactHelpers/PointOfInterestModule.{h,m}` and the `ReactNativeAutoPlayPackage` registration are gone. There is a `src/specs/PointOfInterestTemplate.nitro.ts` with a `nitro.json` autolinking entry and a `HybridPointOfInterestTemplate` on both platforms. The template itself now extends the shared base on each side — `template/PointOfInterestTemplate.kt extends AndroidAutoTemplate` (registered in `AndroidAutoScreen.applyConfigUpdate`) and `ios/templates/PointOfInterestTemplate.swift extends AutoPlayTemplate` (registered in `TemplateStore`) — so it pushes/pops through the shared stack (`Template.push()`, `HybridAutoPlay.popTemplate()`) instead of carrying its own `push`/`pop` bridge methods. `title` is an `AutoText`, `status` is a union that codegens to a native enum, and `autoDismissMs`/`onPopped`/`onWillAppear`… come from the shared `TemplateConfig`.

**`DeviceEventEmitter`/`RCTDeviceEventEmitter` are gone.** Both platforms deliver callbacks through the config now, which is how every other template does it (`onPopped` already lived in `TemplateConfig`). I did not add `addListenerPopped`/`addListenerSelectItem` methods with cleanup callbacks — that would have been a second mechanism for something the base class already provides. Happy to switch if you prefer listeners.

**`onPopped` actually fires now** (the automated review's finding). Android: `AndroidAutoScreen` `ON_DESTROY` → `AndroidAutoTemplate.onPopped()` → `config.onPopped()`. iOS: `TemplateStore.removeTemplate` → `template.onPopped()` → `config.onPopped()`. That covers both app-initiated pops and the native back button.

**`CarToast.makeText` is gone** and the on-press paths are wired through: the action strip is `actionStrip: { label, onPress(template) }` (Android-only — `CPPointOfInterestTemplate` has no action strip), and row press / POI selection both call `onSelectItem(template, itemId)`.

**Colors are configurable**: `colors: { available, busy, inactive, highlight }` as `ThemedColor | string`, defaults are the previous values. The pin cache key includes the resolved colors. White/black stay hardcoded for the count label and the inactive cross — those are contrast colors against the pin color rather than brand colors, but say the word and they go into the config too.

One behaviour fix that came out of this: the old ObjC renderer ignored `isHighlighted`, so the highlight ring only existed on Android. Both platforms draw it now.

**Verified here:** `tsc --noEmit`, `biome check src` and `dpdm` are clean, and `npx nitrogen` regenerates the committed output byte-identically. **Not verified:** I have no Android SDK or Xcode on this machine, so the Kotlin and Swift are review-only — I checked the Car App / CarPlay API signatures against the SDK headers and the nitrogen-generated specs, but a build on your side is the real check.

PR #114, opened 2026-08-25
fix: Fabric surface touches on the virtual display never reach JS

forki:
## What

`SurfaceCallback.onClick` only carries tap coordinates from the host — it doesn't dispatch a real touch event into the Fabric surface running on the virtual display. Only the config-level `onClick` callback (unchanged, still fires) ever ran; RN `Pressable`/`TouchableOpacity` components rendered inside the map surface never received a touch and never ran their own press handlers.

## Why / how

Two things were needed:

1. Dispatch a synthetic `ACTION_DOWN`/`ACTION_UP` `MotionEvent` into the `ReactSurfaceView` from `onClick`, converting the host's surface coordinates to `ReactSurfaceView`-local coordinates via the render scale already computed in `initRenderer`.
2. `ReactSurfaceImpl` silently no-ops `dispatchTouchEvent` until `attach(ReactHost)` has been called on it — `ReactSurfaceView` alone isn't enough. The bridgeless `ReactContext` holds its `ReactHostImpl` as a private field with no public accessor, so it's fetched via reflection in `FabricMapPresentation.onCreate` and attached explicitly. Calling `attach()` again on an already-attached surface throws `IllegalStateException`, which happens on every `onCreate` after the first (e.g. a resize) and is caught and ignored — not a real error.

## Testing

Verified manually on a real Android Auto head unit: a `Pressable` rendered inside a map-hosted RN surface now receives touches and fires its `onPress`, across cold start and post-resize re-creation. No existing unit test coverage for `VirtualRenderer` in this repo.

### DIFF
--- a/packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/VirtualRenderer.kt
+++ b/packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/VirtualRenderer.kt
@@ -7,9 +7,12 @@ import android.graphics.Rect
 import android.hardware.display.DisplayManager
 import android.hardware.display.VirtualDisplay
 import android.os.Bundle
+import android.os.SystemClock
+import android.util.Log
 import android.view.ContextThemeWrapper
 import android.view.Display
 import android.view.LayoutInflater
+import android.view.MotionEvent
 import android.view.View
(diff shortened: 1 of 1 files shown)


g4rb4g3:
This is by design. You should rather use templates to let the user interact with the app and not render react components. I can imagine Google not approving Android Auto apps doing this.
If you need gestures set up the callbacks on the `MapTemplate` like `onClick`.

PR #115, opened 2026-08-25
fix(ios): tint images at pixel size instead of point size

forki:
Rebased onto current master (0.6.0) and reduced to the one fix that holds up.

## Kept: tint at pixel size

`Parser.getTintedImage` built its `CGContext` from `uiImage.size` — the *point* size. A 128px asset at scale 4 therefore tinted into a 32x32 context and came back with a 4x smaller point size than the untinted asset, so tinted icons rendered smaller and blurry. Per your suggestion this is now fixed where the bug actually is: use the `cgImage` pixel dimensions for the clip rect and the context, and keep the original scale on the result. No hardcoded scale factor.

## Dropped

Thanks for the review — I removed the four bits you asked about:

- **MapTemplate.kt** — reverted. You're right, `Action.APP_ICON` as default is the correct behavior; developers define their actions properly.
- **SymbolFont.kt** — reverted. expo-font uses `res/font/` when configured as documented, so the `assets/fonts/` fallback is unnecessary.
- **CarPlayTemplateExtensions.swift** — reverted, `id` stays force-unwrapped.
- **AutoPlayScene.swift** (`backgroundColor = .clear` / `isOpaque = false`) — reverted. I could not reproduce a difference on the CarPlay simulator either, and I have no reliable before/after evidence to offer, so it should not go in on my word alone. If it turns out to matter I will come back with a reproducer.

The branch is renamed in spirit only (still `feature/misc-fixes`); the title now matches the single remaining change.


packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/template/MapTemplate.kt
```diff
@@ -79,7 +89,12 @@ class MapTemplate(
...
-                        .addAction(Action.APP_ICON)
+                        .addAction(
+                            Action.Builder()
+                                .setIcon(transparentIcon)
+                                .setOnClickListener {}
+                                .build()
+                        )
                         .build()
                 )
             }
```
g4rb4g3: Developers have to define their actions properly, having Action.APP_ICON as default was just to prevent the app from crashing. I don't see a benefit in having an empty button over the Action.APP_ICON.
Please revert that bit, thanks!

packages/react-native-autoplay/ios/extensions/CarPlayTemplateExtensions.swift
```diff
@@ -19,7 +19,10 @@ extension CPTemplate {
         initTemplate(template: self, id: id)
     }
     @objc var id: String {
-        return (self.userInfo as? [String: Any])?["id"] as! String
+        // Force-unwrapping here crashed for any CPTemplate created without going through the
+        // convenience init above (e.g. a raw CPTemplate subclass), which had no "id" key set
+        // at all. Fall back to an empty string instead of crashing.
+        return (self.userInfo as? [String: Any])?["id"] as? String ?? ""
```
g4rb4g3: The id is essential for this lib, if templates are not set up correctly it is fine to crash as only developers will face this.
Please revert that bit, Thanks!

packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/utils/SymbolFont.kt
```diff
@@ -22,8 +22,12 @@ object SymbolFont {
...
             fontName.lowercase(), "font", context.packageName
         )
-        if (id == 0) return null
-        val tf = ResourcesCompat.getFont(context, id) ?: return null
+        // Fall back to assets/fonts/<name>.ttf for apps that load fonts at runtime (e.g. Expo's
+        // expo-font, which copies font files into assets rather than res/font/) instead of
+        // bundling them as a res/font/ resource.
+        val tf = (if (id != 0) ResourcesCompat.getFont(context, id) else null)
+            ?: try { Typeface.createFromAsset(context.assets, "fonts/$fontName.ttf") } catch (_: Exception) { null }
+            ?: return null
```
g4rb4g3: /res/font/font.ttf is exactly what epxo-font uses when setting it up the recommended way and not load it at runtime. We have this on our readme already and wanna stick to that, no need for this then.
https://docs.expo.dev/versions/latest/sdk/font/#configuration-in-app-config

packages/react-native-autoplay/ios/scenes/AutoPlayScene.swift
```diff
@@ -84,6 +84,14 @@ class AutoPlayScene: UIResponder {
...
         }
 
+        // Without isOpaque = false, the compositor still treats the window as opaque even
+        // with backgroundColor = .clear, so CarPlay's native template chrome (action bars,
+        // buttons) gets obscured by the window's default black background instead of showing
+        // through to whatever the host draws behind the RN surface.
+        rootView.backgroundColor = .clear
+        window.backgroundColor = .clear
+        window.isOpaque = false
+
```
g4rb4g3: I don't understand this one, can you show some screenshots on what it fixes? Testing it on the CarPlay Simulator does not show any difference to me.

packages/react-native-autoplay/ios/templates/Parser.swift
```diff
@@ -812,11 +812,15 @@ class Parser {
...
-        )
+        // Tint at the image's original pixel size (scale=1) so getTintedImage renders into a
+        // full-resolution context, then reassign scale=4.0 so the tinted image displays at the
+        // same point size the untinted asset would have. Tinting through the asset's own
+        // (often much lower) scale factor produced visibly blurry/resampled glyph buttons.
+        let tinted = applyTint(uiImage: uiImage, color: assetImage.color, traitCollection: traitCollection)
+        guard let tintedImage = tinted, let cgImage = tintedImage.cgImage else {
+            return tinted
+        }
+        return UIImage(cgImage: cgImage, scale: 4.0, orientation: tintedImage.imageOrientation)
```
g4rb4g3: Having a hardcoded 4.0 scale feels wrong, the proper fix seems to be setting height and width from `cgImage` instead of `uiImage.size` in `getTintedImage`, can you double check this please? Thanks!

g4rb4g3 (changes requested):

PR #117, opened 2026-08-26
feat(ios): optionally keep the pan button while the panning interface is shown

crockalet:
While the panning interface is shown, the pan button is filtered out because only two map buttons fit, which leaves the header back button as the only way out of pan mode. There is no JS API for `showPanningInterface` / `dismissPanningInterface` either, so an app that would rather have the pan button itself toggle panning cannot arrange it.

This adds `keepPanButtonWhilePanning` (iOS, default `false`). When set, the pan button is kept first and the remaining buttons are truncated to the two that fit, which makes the existing toggle in `onPanButtonPress` reachable.

Default off, so existing behaviour is unchanged — I kept it opt-in deliberately rather than flipping it, since the current behaviour is a documented choice in `_invalidate`.

```ts
mapTemplate = new MapTemplate({
  keepPanButtonWhilePanning: true,
  mapButtons: [{ type: 'pan' }, { type: 'custom', ... }],
})
```

Bindings were regenerated with `yarn prepare`, so the C++/Swift/Kotlin/JNI sides are all consistent.

Verified: `typecheck:auto-play` and `lint:auto-play` pass. I have not run the example app on a head unit — happy to if you'd like it confirmed there first.


g4rb4g3:
Thanks for the PR! I see the point on the pan button acting as a toggle but then you pretty much have 2 buttons doing the same thing. The idea was actually to put zoom in and out buttons on the map and exit panning mode with the back  button.
Also putting pan mode on first position might make it swap the position from what the user initially pressed (happens on the example app) which feels odd. That is more of a developer thing to pay attention too though.
Can you share some details on your use case so I can get a better idea?

crockalet:
> Thanks for the PR! I see the point on the pan button acting as a toggle but then you pretty much have 2 buttons doing the same thing. The idea was actually to put zoom in and out buttons on the map and exit panning mode with the back button. Also putting pan mode on first position might make it swap the position from what the user initially pressed (happens on the example app) which feels odd. That is more of a developer thing to pay attention too though. Can you share some details on your use case so I can get a better idea?

Hmm, after further investigation I've found the original reason we had to add this patch was a skill issue on my part. Closing as I don't think there's any need for this anymore, thank you for your time and this lib!

PR #118, opened 2026-08-27
docs: rename CLAUDE.md to AGENTS.md and refresh contents

SamuelBrucksch:
## Summary

- Renamed `CLAUDE.md` to `AGENTS.md` (the tool-agnostic convention) and left `CLAUDE.md` as a symlink so Claude Code keeps working.
- Fixed the parts of the doc that had drifted from the code:
  - Voice input moved off `HybridAutoPlay` into its own `Voice` module / `HybridVoice` wrapper, with an options object, STT mode, chunk streaming, encodings and start/end sounds.
  - The Android headless task is auto-registered on import; there is no `registerHeadlessTask(taskName, component)`.
  - `popToRootTemplate()` lives on `HybridAutoPlay`, not on `Template`.
  - No icon font is bundled any more — apps call `setIconFont` and augment `AutoPlayGlyphMap`; `yarn prepare` does not do font bundling.
  - `nitro.json` lists 14 modules, `biome.json` lives at the repo root, and type names were corrected (`Maneuver`, `useVoiceInput` returning `{ coordinates, query }`).
- Added a "Non-obvious things / gotchas" section covering only what is *not* visible from the file layout or API names, including:
  - the reflection-based `getRootViewForAutoplay` AppDelegate hook that fails silently when missing, and the string-named iOS scene delegates;
  - templates being backed by module-level singletons, stale-id `templateNotFound` errors, the hard-coded `MapTemplate` id, and `MessageTemplate` not extending `Template`;
  - platform-branching action configs and null-on-the-wrong-platform hybrids that no-op silently;
  - the `ReactNativeAutoPlay_*` Gradle properties that swap the Android manifest and sourceSet;
  - committed `nitrogen/generated/` output plus the tag-driven release flow;
  - why the `react-native` (JS timers during CarPlay with the screen off) and `expo-splash-screen` (multiple root views) patches exist.

Every claim was verified against the source rather than carried over from the previous version of the doc.

#### Test plan

- [x] `cat CLAUDE.md` resolves through the symlink and shows the AGENTS.md content
- [x] Docs-only change; no build, lint or runtime impact

Generated with [Devin](https://devin.ai)


SamuelBrucksch:
@claude please review

g4rb4g3 (approved):

PR #120, opened 2026-08-31
feat(android): native backdrop hook under the MapTemplate surface

spendres:
## Why

On Android the React content of a `MapTemplate` is rendered onto the car screen through
a virtual display. Some native views cannot be hosted there as React Native views —
Fragment-based map SDK wrappers are bound to the phone `Activity`, so mounting one on the
virtual display's presentation fails. On iOS a host can already put anything it likes
under the React content via the `getRootViewForAutoplay` AppDelegate hook; Android has no
counterpart.

This PR adds that counterpart: a host-provided native `View` rendered **under** the React
surface of the root display, so a native map (or any other native view) can sit beneath
the app's React overlay. Discussed in #119.

## What

- **`NativeBackdrop` interface** (`view`, `onColorSchemeChanged(dark)`, `destroy()`) and
  **`NativeBackdropRegistry.factory: ((CarContext) -> NativeBackdrop)?`**, registered by
  the host in `Application.onCreate` before the `CarAppService` can start.
- **Per-presentation host / teardown in `VirtualRenderer`.** The factory is consulted once
  per presentation (so again after every surface resize), root display only — clusters
  never get one. The view is added as the presentation root's first child, match-parent,
  under the `ReactSurfaceView`. The previous backdrop is parked alongside the display it
  drew on and destroyed in the same sweep as `pendingDisplays`; `stop()` destroys both the
  current and any parked backdrops so a stop during a resize cannot leak one.
- **Fault isolation.** A factory that throws is logged (`Log.w`) and ignored; the React
  surface still renders.
- **Re-parent hardening.** A view the host hands out more than once is detached from its
  previous parent before being added, instead of crashing `Presentation.onCreate`.
- **Transparent surface view while attached.** `ReactSurfaceView` is painted opaque
  `DKGRAY` at construction and survives resizes by re-parenting, so the background is now
  (re)applied per presentation: `TRANSPARENT` while a backdrop is attached, `DKGRAY`
  otherwise.
- **Color-scheme forward.** `AndroidAutoSession.onCarConfigurationChanged` forwards
  `carContext.isDarkMode` to the current root backdrop *before* its template-specific
  early returns, so the backdrop follows the car's day/night (car app quality MR-1)
  whatever the root template currently is (e.g. a pre-trip `MessageTemplate`).
- **README:** new "Native backdrop under the MapTemplate surface" subsection in the
  Android Auto setup section — interface, registration point, lifecycle contract.

No TypeScript / Nitro spec changes; no iOS changes; no new dependencies
(`androidx.car.app.CarContext` is already on the library's classpath).

## How verified

- Compiled in a real Expo SDK 57 app (RN 0.86, New Architecture) via Gradle
  `assembleRelease`, with a host `NativeBackdrop` wrapping a Fragment-free native map view
  (Google Navigation SDK `NavigationViewForAuto`) registered from `Application.onCreate`.
- Desktop Head Unit, debuggable build, simulated drive — all passed:
  - [x] 1. Backdrop paints under the maneuver/ETA cards at the correct scale (presentation
        built on the `ReactContext`, backdrop on the `CarContext` — no density/letterbox
        mismatch on an 800×400 @160 dpi surface).
  - [x] 2. Simulated drive: backdrop animates; React overlay and the host-drawn template
        (maneuver cards, lanes, ETA) unchanged.
  - [x] 3. DHU `night` / `day` mid-drive → `onColorSchemeChanged` delivered both ways
        (verified via the host's own log), including while the root template is not a
        `MapTemplate`.
  - [x] 4. Home the phone app and reopen → no fatal, backdrop persists (not activity-bound).
  - [x] 5. Phone screen locked ≥ 2 min → backdrop still animating.
  - [x] 6. DHU disconnect/reconnect → previous backdrop destroyed in the sweep, fresh one
        created for the new presentation; `logcat` free of leaked-window /
        `IllegalStateException` noise. (Two consecutive resizes and a `dumpsys meminfo`
        growth check were not exercised on the DHU.)
  - [x] 7. Surface shows the backdrop wherever the React tree is transparent — no stray
        opaque background.

## Compatibility

No behavior change when no factory is registered (the default). On that path the only
diff is `reactSurfaceView.setBackgroundColor(Color.DKGRAY)` being re-applied per
presentation, which equals the value the view already has. Clusters are untouched
(`isCluster` short-circuits the factory lookup). The color-scheme forward is a no-op
when `currentBackdrop` is null.

One observation from the same integration, not part of this PR: on Android the
`AssetImage` `scale`/`width`/`height` fields are not applied — `parseAssetImage` hands
the decoded bitmap straight to `CarIcon`, so a host must supply bitmaps at the car
surface's density itself. Happy to open a separate issue if useful.

### DIFF
--- a/packages/react-native-autoplay/README.md
+++ b/packages/react-native-autoplay/README.md
@@ -203,6 +203,40 @@ In case you have ProGuard enabled (`def enableProguardInReleaseBuilds = true` in
 -keep class com.margelo.nitro.swe.iternio.reactnativeautoplay.** { *; }
 ```
 
+#### Native backdrop under the MapTemplate surface
+On Android the React content of a `MapTemplate` is rendered onto the car screen through a virtual display. Some native views cannot be hosted there as React Native views — Fragment-based map SDK wrappers, for example, are bound to the phone `Activity`. For those the library can place a host-provided native `View` **under** the React surface of the root display: the Android counterpart of the iOS `getRootViewForAutoplay` hook above. Your React tree then draws on top of it as an overlay. This applies to the root display only, not to cluster displays.
+
+```kotlin
+interface NativeBackdrop {
+    /** Added as the presentation root's FIRST child, match-parent. */
+    val view: View
+
+    /** The car's day/night changed (CarContext.isDarkMode) — redraw accordingly. */
+    fun onColorSchemeChanged(dark: Boolean)
+
+    /** Release everything; must be idempotent. */
+    fun destroy()
+}
+
+object NativeBackdropRegistry {
+    @Volatile
+    var factory: ((CarContext) -> NativeBackdrop)? = null
+}
+```
+
+Register the factory in your `Application.onCreate`, before the `CarAppService` can start:
+
+```kotlin
+NativeBackdropRegistry.factory = { carContext -> MyMapBackdrop(carContext) }
+```
+
+Lifecycle contract:
+
+-   The factory is consulted once per presentation — i.e. again after every surface resize — and each backdrop is destroyed when its presentation is replaced or the renderer stops.
+-   A factory that throws is logged and ignored; the React surface still renders.
+-   The React surface view is transparent only while a backdrop is attached. Without a registered factory nothing changes: the surface stays opaque as before.
+-   `onColorSchemeChanged(dark)` is forwarded from `Session.onCarConfigurationChanged`, so the backdrop can follow the car's day/night setting (car app quality guideline MR-1). It fires regardless of which template is currently on screen.
+
 ### Android Auto Customization
 You can customize certain behaviors of the library on Android Auto by setting properties in your app's `android/gradle.properties` file.
 
--- a/packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/AndroidAutoSession.kt
+++ b/packages/react-native-autoplay/android/src/main/java/com/margelo/nitro/swe/iternio/reactnativeautoplay/AndroidAutoSession.kt
@@ -113,6 +113,13 @@ class AndroidAutoSession(sessionInfo: SessionInfo) :
     override fun onCarConfigurationChanged(configuration: Configuration) {
         val colorScheme = if (carContext.isDarkMode) ColorScheme.DARK else ColorScheme.LIGHT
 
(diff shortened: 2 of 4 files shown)


SamuelBrucksch:
@claude please review

g4rb4g3:
Interesting stuff, getRootViewForAutoplay was never intended for such a use case. 🙂  
From what I can see this works fine, I did only a basic test with some view rendering a yellow square though.
One thing got me wondering, why did you skip the cluster? Imho this would be quite nice as most of the time you would render a map on that screen too. What do you think about it?
Ah and please check out the comment claude did. Thank you!

spendres:
Thanks for testing it — and fair point on `getRootViewForAutoplay`; this is the Android answer to the same need (a native map the RN layer can't host on the virtual display).

Two commits pushed:

1. **`destroy()` guarded** — both call sites now go through a `runCatching` helper that logs at WARN, same pattern as the factory. A throwing host `destroy()` no longer skips the `VirtualDisplay` release or the surface teardown in `stop()`, nor the rest of the parked sweep.

2. **Cluster support** — you're right that there was no technical reason to skip it; our host just has a single Nav SDK map view. Rather than have the library decide, the factory now receives the display kind and may return `null`:

   ```kotlin
   enum class NativeBackdropDisplay { ROOT, CLUSTER }
   var factory: ((CarContext, NativeBackdropDisplay) -> NativeBackdrop?)? = null
   ```

   `VirtualRenderer` consults it for every display, passing `ROOT` or `CLUSTER`, and each session forwards day/night to its own renderer (root or cluster) via its module name, before the template-specific early returns. README updated with the signature and a root-only example. Since the hook is unreleased I changed the signature rather than adding an overload — happy to add a `(CarContext) -> NativeBackdrop` convenience if you'd prefer to keep the old shape.

Verification: our host (root-only, returning `null` for `CLUSTER`) is updated to the new signature and its guard tests pass; I haven't been able to exercise a cluster backdrop on a head unit that exposes one, so the cluster path is code-reviewed, not driven. If you have a cluster-capable DHU config handy, a yellow square on the cluster would be the test.

g4rb4g3:
Well done, works just fine. Dropping a really cool yellow and red square from native side with some text from react-native.
image

Since nobody appart from you used this I'm fine with the signature update.
Would you mind sharing the app you used this for?

g4rb4g3 (approved):

PR #121, opened 2026-09-01
fix(ios): synchronise shared mutable state to stop cross-thread crashes

SamuelBrucksch:
## Summary

Fixes a data race reported from the field (Sentry `DXQ`, reproduced on 7.1.7 / iPhone17,1 / iOS 26.6): a crash with `KERN_INVALID_ADDRESS` inside `Dictionary.subscript.getter`, at the read of `HybridAutoPlay.safeAreaInsetsListeners`.

Two threads reach the Swift side:

- **main thread** — UIKit/CarPlay delegate callbacks (`viewDidLayoutSubviews` → `safeAreaInsetsDidChange` → `emitSafeAreaInsets`, `templateDidAppear`, `traitCollectionDidChange`, …)
- **JS/Nitro thread** — every hybrid spec method that is a plain `throws` and never hops actors. That is *all* the `addListener*` methods and *all* the `createXTemplate` methods.

The listener registries and stores they share were plain Swift dictionaries with no synchronisation at all. Swift `Dictionary` is not thread-safe, and a read concurrent with a rehash produces exactly the reported signature. `in_foreground: false` on the crash events fits: CarPlay drives layout while the phone app is backgrounded, racing JS-side teardown.

Auditing the rest of the native code turned up the same pattern in five more places, so they are all fixed here.

## Why `NSLock` rather than `@MainActor`

The codebase already has a two-tool concurrency vocabulary and this change stays inside it:

- `NSLock` for state reachable from the JS thread — `ResultBox` and `stopLock` in `VoiceInputManager.swift` are the existing precedent (`ResultBox` is even documented as "safe across concurrent stop() and recognition callbacks")
- `@MainActor` for UIKit/CarPlay-owned objects — `AutoPlayInterfaceController` is fully isolated, and AGENTS.md already states that native template/window mutation is main-thread-only

Both were already correct policy; they just were not applied to the cross-thread registries. `@MainActor` is not an option for the registries themselves: listener registration is a *synchronous* JS call that must return a teardown closure, so hopping actors would mean changing the Nitro signatures to `Promise` and breaking the public API. Locking is the non-breaking fix. The one place a hop *was* the right tool is the cluster description setter, because that path mutates a real CarPlay object.

Note: `NSLocking.withLock` is iOS 16+ and the pod floor is `min_ios_version_supported` (15.1), so new code uses explicit `lock()` / `defer { unlock() }`.

## Changes

| File | Fix |
|---|---|
| `HybridAutoPlay` | `NSLock` over the three listener dicts; `emit*` snapshot then call out. Fixes the reported crash. |
| `TemplateStore` | `NSLock` over `store`; `removeTemplate(s)` remove under the lock and fire `onPopped()` outside it; `traitCollectionDidChange` snapshots first. Replaced read-filter-reassign with `removeValue`. |
| `SceneStore` | `NSLock` over `store`/`renderState`; accessors no longer touch the dicts directly. |
| `HybridCluster` | `NSLock` over all five registries + `eventQueue`; all five `emit*` snapshot then call out; queue drain moved outside the lock. |
| `HybridCarPlayDashboard` | Same, both dicts. |
| `HybridCluster.setAttributedInactiveDescriptionVariants` | Now hops via `Task { @MainActor }` — it was assigning `CPInstrumentClusterController.attributedInactiveDescriptionVariants` off the main thread. No spec change needed since it returns `Void`, so no nitrogen regen. |
| `SymbolFont` | `NSLock` over the font-name cache (written during JS-thread template construction, read from `@MainActor` `_invalidate()`). |
| `VoiceInputManager` | `recordingStart`/`silenceStart` are now consistently under `stopLock`. The tap's silence read-modify-write and `cleanup()`'s nils were outside it, which made the existing lock ineffective for those two fields. |
| `AGENTS.md` | Documents which methods run on which thread, the `NSLock` convention, and the snapshot-then-call-out rule. |

Everywhere a lock is taken, **listeners are snapshotted under the lock and invoked outside it**. Callbacks run JS and can re-enter — `TemplateStore.removeTemplate` → `onPopped` → `VoiceInputTemplate.onDidDisappear` → `removeTemplate` is a real cycle in this codebase. This also fixes a second latent bug: the old `emit*` iterated a collection a callback could mutate mid-iteration.

## Sentry issues this should cover

All figures are last 30 days on `abrp-mobile`. Confirmed iOS-only unless stated.

**Directly confirmed — the crashing frame is a dictionary this PR now locks:**

| Issue | Signature | Events |
|---|---|---|
| DXQ | `emitSafeAreaInsets` → `Dictionary.subscript.getter` | 4,688 |
| QNQ | `Dictionary.subscript.getter`, `TemplateStore` in stack | 57 |
| M4Z | `templateWillAppear` → `TemplateStore.getTemplate` (TemplateStore.swift:17) → `Dictionary.subscript.getter` | 26 |
| E53 | `Dictionary.subscript.getter`, `SceneStore` in stack | 16 |
| EBZ | same, `TemplateStore` | 15 |
| E4R | `didDisconnectInterfaceController` → `Dictionary.subscript.getter` (i.e. `disconnect()` clearing the store) | 14 |
| EMG / ES7 / HQR / FWD / R5P / R3K | same signature, `SceneStore`/`TemplateStore`, long tail | 18 total |
| QXW | `traitCollectionDidChange` → `_NativeDictionary.makeIterator` — iterating `store.values` while it mutates | 1 |
| QQF | `HybridCluster.emitColorScheme` + `hostSetAttributedInactiveVariants` — both cluster bugs in one event | 1 |

Every one of these carries `KERN_INVALID_ADDRESS at 0x8000000000000010`, the same poisoned bucket pointer, which is what a concurrent Swift `Dictionary` access produces.

## Note on `removeTemplate` ordering

`removeTemplate` now removes the entry and *then* calls `onPopped()`, rather than calling it first. JS cannot observe the difference: Nitro dispatches void callbacks via `AsyncJSCallback::callAndForget` → `Dispatcher::runAsync` → `CallInvoker::invokeAsync` (`react-native-nitro-modules/cpp/utils/JSCallback.hpp:112`), so `onPopped` is queued onto the JS thread and runs after the synchronous removal either way. Removing first is simply the cheapest way to keep the callback outside the lock — one acquisition instead of two, and no window where another thread could swap the entry in between.


SamuelBrucksch:
@claude please review

SamuelBrucksch:
Follow-up stacked on this branch: #122 makes `TemplateStore.disconnect()` and `purge()` fire `onPopped()`, which they currently skip. It targets `bugfix/ios-shared-state-data-races`, so **please merge this PR first** — GitHub will then retarget #122 to `master` automatically.

Kept separate because it is a behaviour change (consumer `onPopped` callbacks start running at disconnect) rather than a data-race fix, and this PR should not have to carry that review.

SamuelBrucksch:
On the `stopLock.withLock` availability flag — I checked this and it is a false positive; no change needed.

Foundation declares it as:

```swift
@available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
extension Foundation.NSLocking {
  @_alwaysEmitIntoClient @_disfavoredOverload public func withLock<R>(_ body: () throws -> R) rethrows -> R
}
```

(from `Foundation.swiftmodule/arm64-apple-ios-simulator.swiftinterface` in the current iOS SDK)

`@_alwaysEmitIntoClient` means the body is inlined into the calling binary rather than resolved against the OS Foundation, so it back-deploys. Verified empirically — this typechecks with no availability diagnostic at both `-target arm64-apple-ios15.1-simulator` and `-target arm64-apple-ios14.0-simulator`:

```swift
let l = NSLock()
func probe() -> Int { return l.withLock { 42 } }
```

The `ReactNativeAutoPlay` pod also builds clean at the 15.1 floor, and `withLock` was already used in `VoiceInputManager` (lines 104 and in `cleanup`) before this PR — so if it were gated at iOS 16 the build would already have been failing.

The style-consistency half of the comment is fair, but the split is per-file and deliberate: `VoiceInputManager` already used `withLock`, so the lines added there match their surroundings, while the newly introduced locks use `lock()` / `defer { unlock() }`. Rewriting six call sites either way would be churn with no functional difference.

The origin of the confusion was mine: an earlier revision of this PR added an AGENTS.md bullet asserting `withLock` is iOS 16+. That was wrong and is corrected in `31d7d3c`, which also records that this has now been flagged in review once so it does not get "fixed" again later.

g4rb4g3 (approved):

PR #122, opened 2026-09-02
fix(ios): fire onPopped when the template store is cleared

SamuelBrucksch:
> [!IMPORTANT]
> **Merge after #121.** This is stacked on `bugfix/ios-shared-state-data-races` and targets that branch, so the diff shown here is just this change. Once #121 merges, GitHub retargets this to `master` automatically.

## Summary

`TemplateStore.disconnect()` and `TemplateStore.purge()` drop templates without calling `onPopped()`. Every other removal path (`removeTemplate`, `removeTemplates`) notifies, so these two are the odd ones out.

Why that is a bug and not just a gap:

- **The documented contract says otherwise.** `TemplateConfig.onPopped` in `src/templates/Template.ts`: *"Fired when the template was removed from the stack … this callback lets you know the template is gone forever."* After a disconnect they are gone forever.
- **Android already does it.** The session's `ON_DESTROY` cascades to every `AndroidAutoScreen`, which calls `AndroidAutoTemplate.getTemplate(moduleName)?.onPopped()` (`AndroidAutoScreen.kt`). Identical app code therefore cleans up on Android Auto but not on CarPlay.
- **The host app outlives the car.** After `didDisconnect` the phone app keeps running, so anything an app tied to a template's lifetime — event listeners, timers, cached render state — leaks for the rest of the process lifetime.

`purge()` has the same gap for `CPSearchTemplate`s popped by a native CarPlay button we cannot intercept. Those were genuinely removed from the stack, so they should notify too.

## How this surfaced

An ABRP bug report: the CarPlay **Live data** grid came up completely blank after reconnecting, permanently, until the app was force-quit ([Featurebase](https://abrp.featurebase.app/p/6a94c49c3320e15f2629a4ea)).

The app's live data template registers listeners and unsubscribes them in `onPopped`. Because a disconnect never fired it, the listeners of the dead template stayed alive and kept consuming the telemetry updates that were meant to refresh the *visible* template, which therefore never rendered a single button. 

## Changes

`packages/react-native-autoplay/ios/templates/TemplateStore.swift`:

| Method | Change |
|---|---|
| `disconnect()` | Snapshot the templates, clear the store under the lock, then fire `onPopped()` on each outside it. |
| `purge()` | Same shape: remove the search templates under the lock, notify outside it. Replaces the read-filter-reassign. |

Removal happens under the lock and callbacks run outside it, matching the convention #121 establishes (callbacks run JS and can re-enter — `removeTemplate` → `onPopped` → `VoiceInputTemplate.onDidDisappear` → `removeTemplate`).

## Behaviour change to be aware of

Consumer `onPopped` callbacks now run at disconnect, a moment they never ran before. That is the point of the fix, but it means teardown code that was previously skipped will execute.

## Test plan

- [ ] Example app on the CarPlay Simulator: push a template with an `onPopped` handler, disconnect the head unit, confirm the callback fires exactly once
- [ ] Reconnect and confirm templates can be pushed again and `onPopped` is not fired a second time for the old ones
- [ ] Open the search template, dismiss it with the native CarPlay button, confirm `purge()` now fires `onPopped`
- [ ] No regression in the `removeTemplate` → `onPopped` → `removeTemplate` re-entrancy path (voice input)

Generated with [Devin](https://devin.ai)

### DIFF
--- a/packages/react-native-autoplay/ios/templates/TemplateStore.swift
+++ b/packages/react-native-autoplay/ios/templates/TemplateStore.swift
@@ -53,11 +53,18 @@ class TemplateStore {
     }
 
     func purge() {
-        withLock {
-            store = store.filter {
-                !($0.value.getTemplate() is CPSearchTemplate)
+        // These templates were popped by a native CarPlay button we cannot
+        // intercept, so they need an `onPopped()` just like an explicit pop.
+        let removed = withLock {
+            let searchTemplates = store.filter {
+                $0.value.getTemplate() is CPSearchTemplate
             }
+            searchTemplates.keys.forEach { store.removeValue(forKey: $0) }
+
+            return Array(searchTemplates.values)
         }
+
+        removed.forEach { template in template.onPopped() }
     }
 
     @MainActor
@@ -68,6 +75,19 @@ class TemplateStore {
     }
 
     func disconnect() {
-        withLock { store = [:] }
+        // The session is gone, so these templates will never be popped
+        // individually. Without an `onPopped()` here anything the app tied to a
+        // template's lifetime (event listeners, timers, cached render state)
+        // leaks for the rest of the app lifetime, and the host app keeps
+        // running after a disconnect. Android does the same on
+        // `Lifecycle.Event.ON_DESTROY` (AndroidAutoScreen.kt).
+        let removed = withLock {
+            let templates = Array(store.values)
+            store = [:]
+
+            return templates
+        }
+
+        removed.forEach { template in template.onPopped() }
     }
 }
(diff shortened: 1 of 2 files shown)


SamuelBrucksch:
@claude please review

packages/react-native-autoplay/ios/templates/TemplateStore.swift
```diff
@@ -68,6 +75,19 @@ class TemplateStore {
     }
 
     func disconnect() {
-        withLock { store = [:] }
+        // The session is gone, so these templates will never be popped
+        // individually. Without an `onPopped()` here anything the app tied to a
+        // template's lifetime (event listeners, timers, cached render state)
+        // leaks for the rest of the app lifetime, and the host app keeps
+        // running after a disconnect. Android does the same on
+        // `Lifecycle.Event.ON_DESTROY` (AndroidAutoScreen.kt).
```
g4rb4g3:
```suggestion
        // fire onPopped when disconnecting for any template that was still visible
```

g4rb4g3 (approved):

PR #130, opened 2026-09-18
docs: split agent instructions so they survive truncation

SamuelBrucksch:
## Summary

- `AGENTS.md` 24.5 KB → 13 KB, restructured so nothing important gets truncated. Devin cuts at 16 KB, which dropped the whole Code Style section.
- Nitro is now a rule at the top naming the forbidden constructs, not a description 3 KB in. #110 added a `TurboReactPackage` and a `NativeModules` lookup.
- Deep reference moved to `docs/*.md` with a `docs/README.md` index, linked from the root README. It is contributor documentation, not agent-only — `.skills/` (symlinked into `.claude/`, `.agents/` and `.cursor/`) summarises three of them for agents.
- Added a PR template, modelled on the conventions already used in this repo.
- New rules: PR descriptions stay short and high level; a tool opening a PR signs off with the tool **and the model** it ran as.
- Both AI reviewers now flag stale docs and over-long descriptions. The model sign-off is encouraged in the rules only, not enforced.
- `AGENTS.md` is the single rule file: Cursor, Devin and Copilot (coding agent, code review, CLI, VS Code chat) all read it, so there are no tool-specific mirrors to drift.

## Bugs found while verifying every claim against the source

- **README Automotive setup was broken.** It told consumers to put bare `minSdkVersion` / `isAutomotiveApp` in `gradle.properties`. `getExtOrDefault` reads `rootProject.ext.<name>` then `ReactNativeAutoPlay_<name>`, so the unprefixed form is silently ignored and you get an Android Auto build. The example app only ever used the `rootProject.ext` route, so this path was untested.
- **README bundle-identifier steps** told consumers to open `example.xcodeproj` and select the example target.
- **`AGENTS.md` claimed Automotive "needs `minSdk >= 29`".** Nothing enforces it — the host app must raise it.
- **`postOnUiAndAwait`** was documented as covering "all screen-stack mutations". It covers stack ops on `HybridAutoPlay`; all seven `Hybrid*Template` classes use `Promise.async`.
- **A pinned patch version in prose**, stale on the next RN bump.

## Notes for review

- Setup instructions are no longer duplicated — the README is authoritative, the agent docs keep only mechanics and silent failure modes.
- `ai-review/test.js` already fails `npm test` on `master` (no tests in the file). Untouched.

## Test plan

- [x] `yarn lint:auto-play` and `yarn typecheck:auto-play` pass
- [x] `node --check ai-review/ai-review.js`; `claude.yml` parses as YAML
- [x] All links resolve; nine skill symlinks commit as mode `120000`
- [x] Re-verified every doc claim against the post-merge tree
- [ ] N/A — docs and CI config only, no runtime impact

## Checklist

- [x] `yarn lint:auto-play` and `yarn typecheck:auto-play` pass
- [x] `package.json` version is not hand-edited
- [x] Agent-facing docs updated — this PR is that change

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)


SamuelBrucksch:
@claude please review. Make sure no important information is lost and all the information reachable before is still reachable for AI tools.

SamuelBrucksch:
Addressed in 47bd09f. Both issues were real; all four dropped details restored and re-verified against source.

**1. Size** — dropped the self-reported figure rather than correcting it to 13 KB. It goes stale on every edit, which is how it got wrong in the first place. Kept the 16 KB ceiling and added `wc -c AGENTS.md`.

**2. Mirror rule overclaimed** — the sharpest finding: the drift check would have fired on its own PR. Confirmed the mirrors omit seven of the fifteen bold rules. Resolved both ways, since the subset is deliberate but two of the omissions weren't:
- `AGENTS.md` now says a **subset** is mirrored — the rules that apply while editing code or opening a PR — and that session etiquette (don't commit unless asked, no summary files, the sync rule) stays there only.
- Both reviewers now treat **absence as fine** and report only contradictions, or a mirrored rule changed on one side.
- Added the README-update and lint/typecheck rules to both mirrors. Agreed the README one matters most — it's the one that silently breaks consumers.

**Dropped details** — all restored:
- `.android.ts` no-op fallback now documented in `docs/native-modules.md` as a second, distinct pattern from the `null` hybrid object, with the note that it's the one to copy for the next Android-only module.
- Template `id` derivation restored next to the "id is the only correlation key" gotcha it explains.
- `AutoPlayModules` got its own section — worth more than a mention, since cluster ids are *not* in the enum and must be read from `RootComponentInitialProps.id`. Precise member names are `App = 'main'`, `AutoPlayRoot`, `CarPlayDashboard`.
- Provider paths restored.

**On the SKILL.md relative-link note** — good catch and I've left it as-is deliberately. `../../docs/…` resolves correctly through the symlink for anything that hits the filesystem, and correctly on GitHub when browsing `.skills/`. A lexical normaliser landing on `.claude/docs/…` would be a real miss, but the visible link text is the root-relative path and `AGENTS.md` links the same files directly, so an agent that follows the text still gets there. Flagging rather than fixing, since the fix (dropping the hyperlink) costs human navigability.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)

SamuelBrucksch:
@claude check again if the issues you mentioned were resolved properly

